PR #5, opened 2026-04-21
fix: correctly handle email attachments end-to-end

vroomanj:
- Fix inline vs downloadable attachment distinction: only treat an attachment as inline when both Content-ID and Content-Disposition: inline are present; previously any attachment with a Content-ID (common in many email clients) was hidden from the download list
- Show downloadable attachments in the full email modal (EmailHtmlModal), not just in the message preview
- Add paperclip indicator to person list rows for threads that have any attachment across their email history


choyiny (approved):
thank you for your contribution! Appreciate this.

PR #8, opened 2026-04-22
feat: real-time inbox updates via Durable Object WebSockets

vroomanj:
Add a NotificationsHub Durable Object that maintains persistent WebSocket connections and pushes email_received events to connected clients instantly, replacing the need for manual refresh.

- NotificationsHub DO: hibernatable WebSocket connections keyed by inbox, with auth metadata (isAdmin + allowed inboxes) serialized per socket
- /api/notifications/stream: authenticated WebSocket upgrade endpoint; session and inbox permissions validated in the main worker before forwarding to the DO
- email-handler: notifies the hub (best-effort, via ctx.waitUntil) after each inbound email is inserted
- useRealtimeUpdates hook: opens a WebSocket, reconnects on close, triggers a callback on email_received events
- InboxPage, PersonList, PersonDetail: wired to refreshKey so both the person list and open thread refetch automatically on new mail
- wrangler.jsonc.example: documents the DO binding and v1 migration for fresh deployers


vroomanj:
I'm not entirely sure what Claude did with worker-configuration.d.ts

I'm trying to look now.

vroomanj:
Actually, I think that file (worker-configuration.d.ts) should be git ignored???

choyiny (changes requested):
Thanks for the contribution! I'm really happy with the overall DO WebSocket architecture. I just have a couple of points to clarify before we move forward:

Permission Scoping: Are the WebSocket events properly scoped by permissions? It looks like the system currently lacks permission checks here. We need to ensure that regular users only receive events for inboxes provisioned to them, while admins retain visibility across all inboxes.

worker-configuration.d.ts: Since this file is automatically generated by running `wrangler types`, I think we should keep it in the codebase rather than removing or ignoring it.

vroomanj:
The Permission Scoping should be handled already. Please bear with me here as I am relying on Claude a lot here. I will just show you what Claude had to say in regards to your concern:

> ● The permission scoping is actually already implemented correctly — let me walk through   
>  exactly how it works and identify the one genuine gap.                                   
>                                                                                           
>  What's already working:                                                                  
>                                                                                           
>  1. /api/notifications/stream runs through the full auth middleware stack (session        
>  validation → injectAllowedInboxes) before reaching the router                            
>  2. The router reads the resolved allowedInboxes and passes it to the DO as headers       
>  3. The DO stores { isAdmin, inboxes } per WebSocket connection using serializeAttachment 
>  4. At broadcast time, the DO checks meta.isAdmin || meta.inboxes.includes(inbox) before  
>  sending — so regular users only get events for their inboxes, admins get everything      
>                                                                                           
>  The one real gap — stale permissions:                                                    
>                  
>  Inbox permissions are captured at connection time. If an admin revokes a user's inbox    
>  access after they've connected, that user's WebSocket will keep receiving events for the
>  revoked inbox until they reconnect (e.g. page refresh).                                  
>                  
>  The fix is to re-validate permissions on each broadcast by calling back to D1 — but that 
>  adds a DB query per inbound email per connected user, which is expensive. The better
>  tradeoff is to keep the current approach but force a reconnect when permissions change.  
>                  
>  Is the stale-permissions case what the repo owner is concerned about, or is there        
>  something more specific they flagged?

Want me to fix that stale-permissions gap first?

choyiny:
This makes sense - I was reading the global DO that is created and it loops through the list of active websocket connections and filters by the incoming header for the inboxes.

I've been rethinking the architecture and I think that DOs should be scoped per user instead of having one 'global'. When an email comes in, then, we will find the users that will be recipients (by inbox_permissions & their admin role), and then make a call to the user's DO. The DO then is responsible for sending the websocket message (same as now)

The reason why I want to change it to this is to account for more notification methods. Browser push notifications, SMS notifications are something that I'd like for this solution to have. For push notifications, it is just feature parity with Gmail, but for support I'm thinking a whatsapp/sms integration would be helpful.

Then the `X-Is-Admin` / `X-Allowed-Inboxes` headers can be removed.

Also, found a security issue with the websocket upgrade endpoint:

Cross-Site WebSocket Hijacking (CSWSH)
```
worker/src/routers/notifications-router.ts:9 does not validate the Origin header before accepting the upgrade. 
WebSockets are not subject to the Same-Origin Policy or CORS, and the browser will attach the BetterAuth session
cookie to a cross-site WS handshake. A malicious page the victim visits can ope
wss://<your-saasmail>/api/notifications/stream, authenticate as the victim, and receive email_received events
(inbox addresses + arrival timing) for every inbox the victim is allowed on.

The fix is an explicit allow-list check against c.req.header("Origin") using TRUSTED_ORIGINS. Leaving it off is one
of the most common mistakes when adding WebSockets to cookie-auth apps.
```

Finally, I didn't do a good job at running wrangler types before - I've updated that in the main branch. Would you mind rebasing on main?

Happy to do the changes as well - let me know :)

.gitignore
```diff
@@ -16,3 +16,4 @@ playwright-report/
 
 # Local superpowers docs (specs, plans)
 docs/superpowers/
+worker-configuration.d.ts
```
choyiny: remove this - also, does running `wrangler types` make it not diff that much? Curious

vroomanj: Sorry, I didn't mean to add that back in.

worker-configuration.d.ts
vroomanj, line 1: I can't get this to not be a huge diff? I'm not sure what I'm doing wrong. I have ran `wrangler types`

choyiny: No problem! I'll see if I can fix this and also address some nitpicks.

choyiny (approved):
lgtm!

PR #10, opened 2026-04-22
fix(email-sender): build MIME for Cloudflare send_email to pass custom headers

michaelmcnees:
The Cloudflare Email Sending binding's builder form (env.EMAIL.send({from, to, subject, html, headers})) rejects any non-whitelisted custom header with:

  "custom header 'Message-ID' is not allowed. Only whitelisted headers
   and X-* headers are accepted."

saasmail always attaches Message-ID (and In-Reply-To on replies) for Gmail/Outlook threading, so every outbound email over the Cloudflare provider currently fails. The Resend provider is unaffected.

Rewrite CloudflareSender.send() to construct a raw MIME message via mimetext.createMimeMessage() and hand it to new EmailMessage(from, to, raw) — the raw-MIME path accepts arbitrary headers. Added a parseFrom() helper that splits "Name <addr>" into name/address for EmailMessage's envelope sender, and added mimetext as a runtime dependency. The Resend path is unchanged.

Repro: configure the EMAIL binding (no RESEND_API_KEY) and call any endpoint that sends mail — the binding throws the header-rejection error above. With this change, Message-ID / In-Reply-To ride inside the MIME body and the send succeeds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "drizzle-orm": "^0.45.2",
     "hono": "^4.12.14",
     "lucide-react": "^1.8.0",
+    "mimetext": "^3.0.28",
     "nanoid": "^5.1.5",
     "postal-mime": "^2.4.1",
     "react": "^18.3.1",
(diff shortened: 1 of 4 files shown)


choyiny (commented):
I somehow missed this during a refactor - thank you for your contribution!

choyiny (approved):
I somehow missed this during a refactor - thank you for your contribution!

PR #23, opened 2026-04-23
feat: full-text email search via FTS5

vroomanj:
## Summary

- Adds an `emails_fts` FTS5 virtual table backed by the `emails` table, indexing `subject` and `body_text`
- INSERT/UPDATE/DELETE triggers keep the index in sync automatically; existing emails are backfilled on migration
- The grouped people endpoint ORs FTS matches into the search condition — typing in the search box now surfaces people whose **emails mention the query**, not just people whose name/email match
- Search box placeholder updated from "Search people..." to "Search..."
- Search box gains a clear (×) button that appears when there is text

## Why FTS5

LIKE-based search requires a full table scan. FTS5 uses an inverted index so only matching rows are touched — more efficient at scale and no extra D1 cost per query.

## Test plan

- [x] Search for a word that appears in an email subject or body — the sender should appear in results even if their name/email don't match
- [x] Verify the × button clears the search box
- [x] Confirm existing name/email search still works
- [x] Run `yarn db:migrate:prod` (or dev) to apply `0019_emails_fts.sql` before deploying

### DIFF
--- a/migrations/0020_emails_fts.sql
+++ b/migrations/0020_emails_fts.sql
@@ -1,0 +1,30 @@
+-- FTS5 virtual table for full-text search across email subject and body.
+-- Uses emails as a content table so text is not duplicated on disk.
+-- Triggers keep the index in sync with the emails table.
+CREATE VIRTUAL TABLE emails_fts USING fts5(
+  subject,
+  body_text,
+  content='emails',
+  content_rowid='rowid'
+);
+
+-- Backfill existing emails into the index
+INSERT INTO emails_fts(rowid, subject, body_text)
+SELECT rowid, subject, body_text FROM emails;
+
+CREATE TRIGGER emails_fts_ai AFTER INSERT ON emails BEGIN
+  INSERT INTO emails_fts(rowid, subject, body_text)
+  VALUES (new.rowid, new.subject, new.body_text);
+END;
+
+CREATE TRIGGER emails_fts_ad AFTER DELETE ON emails BEGIN
+  INSERT INTO emails_fts(emails_fts, rowid, subject, body_text)
+  VALUES ('delete', old.rowid, old.subject, old.body_text);
+END;
+
+CREATE TRIGGER emails_fts_au AFTER UPDATE ON emails BEGIN
+  INSERT INTO emails_fts(emails_fts, rowid, subject, body_text)
+  VALUES ('delete', old.rowid, old.subject, old.body_text);
+  INSERT INTO emails_fts(rowid, subject, body_text)
+  VALUES (new.rowid, new.subject, new.body_text);
+END;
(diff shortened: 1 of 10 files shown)


migrations/0020_emails_fts.sql
```diff
@@ -0,0 +1,30 @@
+-- FTS5 virtual table for full-text search across email subject and body.
+-- Uses emails as a content table so text is not duplicated on disk.
+-- Triggers keep the index in sync with the emails table.
+CREATE VIRTUAL TABLE emails_fts USING fts5(
```
choyiny, line 4: I understand this is Claude generated - can we redo this by asking drizzle to generate a manual migration then paste this in? The missing piece would be the meta/0019_snapshot.json that it autogenerates. Also - timestamps in _journal.json is hallucinated

choyiny (changes requested):
having vitests and e2e playwright tests would be much appreciated here - in my experience, that's also 1 claude prompt away.

migrations/0020_emails_fts.sql
vroomanj: I think I took care of this just a bit ago? I had to rebase and reapply some changes. I'm working on vitests and e2e playwright tests now.

worker/src/routers/people-router.ts
```diff
@@ -80,8 +80,14 @@ peopleRouter.openapi(listGroupedPeopleRoute, async (c) => {
...
   if (q) {
     const pattern = `%${escapeLike(q)}%`;
+    const ftsQuery = escapeFts(q);
     conditions.push(
-      sql`(s.email LIKE ${pattern} ESCAPE '\\' OR s.name LIKE ${pattern} ESCAPE '\\')`,
+      sql`(s.email LIKE ${pattern} ESCAPE '\\' OR s.name LIKE ${pattern} ESCAPE '\\'
+        OR s.id IN (
+          SELECT person_id FROM emails
+          JOIN emails_fts ON emails.rowid = emails_fts.rowid
+          WHERE emails_fts MATCH ${ftsQuery}
```
choyiny: feels like there maybe a potential SQL injection problem here, could it?

choyiny (changes requested):

worker/src/routers/people-router.ts
vroomanj: I'm not certain myself but Claude says no anyhow.

> ● No. The ${ftsQuery} is inside Drizzle's sql tagged template literal — values interpolated
>    that way become ? bound parameters in the actual SQL sent to SQLite, not raw string
>   concatenation. You can see this in the test output from earlier:                         
>                            
>   WHERE emails_fts MATCH ?                                                                 
>   -- params: ["%O'Brien%", "%O'Brien%", "\"O'Brien\""]                                     
>                                                                                            
>   The escapeFts function is a second, FTS5-specific layer — it prevents operator injection 
>   (e.g. a user typing AND OR * to exploit FTS5 query syntax), not classic SQL injection.   
>   Both protections are in place.

choyiny: my other concern is the need to scope to member's inbox permissions, since members can be configured to only access certain mailboxes.

choyiny (approved):

choyiny:
thank you for your contribution again!

PR #24, opened 2026-04-23
feat: admin delete-person with kebab menu

vroomanj:
## Summary

- Adds a `DELETE /api/people/:id` endpoint (admin-only, returns 403 for non-admins) that hard-deletes a person and cascades through their R2 attachments, attachment DB records, received emails, sent emails, and the person record itself
- Adds `deletePerson()` to the frontend API client
- Shows a hover kebab (⋯) button on each person row — visible only to admins — with a "Delete person" action and confirmation dialog
- Dropdown is rendered via `createPortal` into `document.body` to escape `ScrollArea`'s overflow clipping
- Non-admin users see no button and no extra right padding on person rows (fixes visual alignment)
- Selecting a deleted person in the right panel is automatically cleared

## Test plan

- [x] As admin: hover a person row → `⋯` appears → click → "Delete person" → confirm → person and all emails removed
- [x] As admin: dismiss the dropdown by clicking outside or scrolling
- [x] As non-admin: no `⋯` button visible, row padding is symmetric
- [ ] `DELETE /api/people/:id` returns 403 for non-admin session
- [ ] `DELETE /api/people/:id` returns 404 for unknown ID
- [x] Deleting the currently-selected person clears the right panel

### DIFF
--- a/e2e/specs/delete-person.spec.ts
+++ b/e2e/specs/delete-person.spec.ts
@@ -72,8 +72,10 @@ test.describe.serial("delete person", () => {
     await expect(bobRow).toBeVisible();
     await bobRow.click();
 
-    // Right panel should show Bob's emails.
-    await expect(page.locator("text=bob@customers.test")).toBeVisible();
+    // Right panel should show Bob's detail heading (Bob Brown is the seeded name).
+    await expect(
+      page.getByRole("heading", { name: "Bob Brown" }),
+    ).toBeVisible();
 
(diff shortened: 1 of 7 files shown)


choyiny (changes requested):
thanks for your continuous contribution! The final step would be to ask Claude to add a vitest and a playwright end to end test to test this behavior, and we're good to merge.

choyiny (approved):

PR #66, opened 2026-05-07
feat(v0.5.0): groups + CC, tray modals, table fixes, design refresh

MahmoudHalat:
## Summary

What started as a "stack CC support onto threads" PR has grown into the v0.5.0 cut. 20 commits, ~16k net lines, no unrelated drive-by changes — every commit either ships a feature, fixes a bug, or refactors a surface to match the design system. Touches the inbox, all three tray modals (compose / reply / view-original), templates, sequences, the admin/users page, branding/settings, and the seed data.

The work groups into seven themes below. Each section lists the commits that contributed and a short summary of what changed and why.

---

## 1. Multi-recipient threads + group conversations

The big functional addition: emails with 2+ external participants now form their own "group conversation" rows in the inbox, each with a single `conversation_id` keyed by `(inbox, sorted-externals)`.

- **CC support end-to-end** — schema migration `0021_email_cc.sql` adds `cc TEXT` to both `emails` and `sent_emails` (JSON-encoded array). Inbound parsing reads `parsed.cc` from postal-mime; outbound write paths accept a `cc[]` array. Reply auto-fills CC from the original (with own inboxes filtered out), Compose has a CC chip input.
- **Conversation IDs** — migration `0022` adds an indexed `conversation_id TEXT` column on both tables, computed via `computeConversationId(inbox, externals) = c_<sha256(...).slice(0,16)>`. Both inbound and outbound paths stamp the column at write time.
- **Group rows in the list** — `/api/people/grouped` returns a discriminated union of person and group rows, sorted/paginated together. List view renders a `GroupRow` with overlapping avatar stack and comma-separated names; table view renders a `GroupTableRow` with the same palette and full bulk-select support.
- **Merged group timeline** — `ConversationDetail` page shows every email from the conversation in one chat- or thread-mode feed, with a `senderResolver` so each bubble gets its actual sender's name in chat mode.
- **Real demo attachments** — `seeds/upload-attachments.sh` ships a half-dozen real PDFs/PNGs/CSVs into local R2; ~15% of inbound emails get one or two.
- **Roster-diff bug fix** — the inline "+ X added · − Y removed" notice between consecutive thread messages was diffing only the `cc[]` arrays. A participant who's the *sender* of a given message naturally isn't in their own CC, so every speaker rotation falsely read as "X removed." `src/lib/roster.ts` now builds the full participant set (`from + to + cc`, deduped) per email so speaker rotations cancel out.

Commits: `5e11a09`, `518c7e0`, `86fa192`, `5bf956b`, `fcc103c`, `aaacf07`, `d1badcd`, `8932ffe` (roster fix bundled with chat→full-compose handoff)

---

## 2. Tray modals — compose / reply / view-original

Compose, Reply, and View-Original were previously full-height right-side drawers that took focus and dimmed the background. Switched all three to a Gmail-style tray family.

- **Bottom-right tray** — anchored at `bottom-0 right-6` on ≥sm; falls back to full-width 90vh on mobile. Slides up from the bottom edge instead of in from the right.
- **Non-modal** — `<DialogPrimitive.Root modal={false}>` keeps the inbox underneath fully clickable, no focus trap, no body-scroll lock. Only the X button or Esc dismisses; `onInteractOutside` is `preventDefault`'d so accidental clicks on the inbox don't lose a draft.
- **Compact chrome + fullscreen toggle** — original chrome was eating ~50% of vertical space (big avatar header + 4-row grid metadata + dual-line footer). Slim h-11 single-row header, dense `TrayMetaRow` fields, slim footer — body went from ~290px to ~500px in compact mode. Each tray gets a `Maximize2` button that grows it to `inset-3 sm:inset-8` for editor-heavy work, restored via `Minimize2`.
- **Rounded-corner bleed fix** — added `overflow-hidden` so child backgrounds stop poking past the rounded top edges.
- **Chat → full-compose handoff** — `ChatQuickReply` gets a small expand icon. Clicking opens `ComposeModal` pre-populated with the reply context (from inbox + recipient + CC roster + Re: subject) so the user can swap sender identity, add a teammate to CC, attach files. Threads `DashboardLayout.openCompose(prefill?)` cleanly through the outlet context.

Commits: `adf9e51`, `ee588fb`, `8932ffe`

---

## 3. Templates / Sequences / Users — design system refresh

These admin surfaces were running on legacy chrome (custom top bars, raw HTML entities for nav, off-system inputs, tiny 10–11px fonts). Brought them into line with `SettingsPage` and the rest of the app.

- **Shared `src/components/PageForm.tsx`** — `Field`, `SectionHeader`, `PaneLabel`, `CodeBlock` (with copy button), `FORM_INPUT_CLASS`. Reused on the rebuilt template editor, sequence editor, and sequence detail.
- **Templates editor** — full rewrite. PageContainer + PageHeader with breadcrumb back-link, contextual `Create template` / `Save changes` buttons. Form lives in cards. Code/Split/Preview view toggle (mirrors the ReplyComposer mode-toggle pattern). Auto-detected `{{variables}}` surface as a violet-chip strip. Replaces the old slide-over `<ApiSamplePanel>` with an inline collapsible "Send via API" card.
- **Templates list** — small variable-count chip per row.
- **Sequences editor** — full rewrite. Steps are a numbered timeline with violet circles, per-step Template + Delay fields, "Day N · sends right away" cumulative hint. Inline error notice replaces an `alert()`. Collapsible "Enroll via API" card with copy-button code blocks.
- **Sequences detail** — connected timeline of steps; enrollments grouped by status (Active / Completed / Cancelled) with avatar circles, status chips, and per-row sent/total progress bars.
- **Sequences list** — violet step-count chip + Clock "spans X days" hint per row.
- **Admin → Users** — drops a dead duplicate `<Dialog>` tree that had been hidden in the markup since a prior refactor. Members get avatar circles using the same opaque palette as group threads, "you" pill on the current user, Passkey + Admin chips with consistent shape, `MoreHorizontal` dropdown trigger replacing the literal `"..."`. Invite dialog: 2-up role chooser cards, copy button gets an icon.

Commits: `b213290`, `d0237b0`, `d858ea6`

---

## 4. Inbox table view — pagination + bi-directional sort + accurate aggregates

Three real bugs surfaced in one go. Architected end-to-end so list and table view share a single contract.

- **No pagination UI in table view.** List view (`PersonList`) had prev/next since day one; `PeopleTable` was rendering whatever happened to be on the current page with no nav. New `<PageFooter>` mirrors `PersonList`'s pattern + a "Showing N–M of K" range.
- **Stat tiles lied past page 1.** "People / Unread / Multi-inbox / With attachments" summed `items.length` — i.e. only the visible page. Server `/api/people/grouped` now returns `aggregates: { unreadRowCount, attachmentRowCount, multiInboxRowCount, totalUnreadEmails }` computed over the merged (pre-paged) result. The Unread tile was extra-wrong: it summed *email* counts across rows while the others counted *rows* — now matches.
- **Sort never toggled direction.** Server `/api/people/grouped` accepts `direction: asc | desc`. Each sort key has a natural default (recency/unread/attachments → desc, inbox → asc). New `InboxSortSpec = { key, direction }` type. Column-header click on the active key flips direction; click on another key switches with the natural default. New explicit ↑/↓ button next to the toolbar's sort dropdown. Localstorage migration handles legacy bare-key persisted values.
- 4 new server tests cover aggregates over multi-page filters, default inbox-asc sort, explicit `direction=desc` flip, and recency-asc oldest-first.

Commits: `416bc8c`

---

## 5. View-original modal — thread context

Per "always make it possible to view a thread mode of all prior messages, even if we're in chat mode."

- `EmailHtmlModal` loads surrounding thread via `fetchPersonEmails(personId, ...)` on open (same call `ReplyComposer` already uses — keeps server load flat).
- New "Show N earlier" pill in the body toolbar. When expanded, prior messages render above the focal email's body in a divided list, with a "This message" divider before the focal HTML. Always available when a thread exists, including in chat-mode parents.
- Extracted `ThreadMessage` from `ReplyComposer` to `src/components/ThreadMessage.tsx` for reuse.

Commits: `3f07477`, `d57c177` (Rules-of-Hooks fix — `useMemo` had been placed after an `if (!email) return null` early-return).

---

## 6. Cross-inbox send awareness

Per "if I'm in 1 chat, and I reply as another chat like billing from sales, it doesn't refresh or tell me that the chat response is in another inbox."

Architected as two custom-event channels (same lightweight pattern as `HIDE_SIGNATURES_EVENT` already used in the codebase — zero context, zero prop-drilling):

- **`src/lib/toast.ts` + `<Toaster>`** — `showToast({ kind, message, description, action, durationMs })` fires `TOAST_EVENT`. A singleton `<Toaster>` mounted in `DashboardLayout` subscribes and renders a stack of dismissible cards bottom-left (away from the bottom-right tray modals).
- **`src/lib/email-events.ts`** — `dispatchEmailSent({ fromAddress, to, origin })` + `onEmailSent(handler)`. `ReplyComposer`, `ComposeModal`, and `ChatQuickReply` all fire on success.
- **`PersonDetail` listens** — refetches; when `fromAddress` is one of the person's inbox tabs and differs from `activeInbox`, auto-switches to that tab and toasts: "Reply landed in `billing`. You were viewing `sales` — switched tabs so you can see it."
- **`ConversationDetail` listens** — group `conversation_id`s are keyed by `(inbox, sorted-externals)`, so a reply from a different inbox starts a new conversation entirely. Refetches + warning toast: "Reply went to `billing`. This thread is on `sales` — replies from another inbox start a new thread."

Server bug fixed in the process: `GET /api/emails/:id` only checked the received-emails table. Clicking Reply on one of our own outgoing messages 404'd, and a silent `.catch(() => {})` in `ReplyComposer` masked it — empty composer with no warning. Now falls back to `sent_emails` with the same defense-in-depth ownership check the reply route uses.

Commits: `3f07477`, `fe39526`

---

## 7. Branding, signatures, settings, demo seed

- **Per-inbox signatures** — `0023_inbox_signatures.sql` adds `signature_html TEXT` on `sender_identities`. Admin Inboxes table grows a "Signature" column with inline TipTap editor. Compose/Reply auto-attach the active inbox's signature wrapped in `<div data-signature>` so the chat-feed "Hide signatures" toggle can strip it back out cleanly.
- **HTML-in-chat fallback** — non-compliant emails (marketing/transactional with broken plain-text fallback) render an inline preview card that opens the View-Original modal.
- **Admin-editable brand name** — `0024_app_settings.sql` adds an `app_settings` KV table. `GET /api/config` returns `brandName` (defaults to "saasmail"); `PATCH /api/admin/settings` lets admins set/reset (1–40 chars, validated). `BrandingProvider` exposes `brandName + refresh()` so the wordmark updates live without a page reload. TopNav, Wordmark, LoginPage, InviteAcceptPage all consume it.
- **Settings page restructure** — three sections now: Notifications / Display preferences (Hide-signatures toggle moved here from the TopNav dropdown) / App branding (admin-only). Replaces the old single-purpose `NotificationsSettingsPage`.
- **Demo seed: templates + sequences** — 7 email templates spanning outreach / onboarding / support / meetings, mixing inbox-scoped and global. 3 sequences (4-step Outreach, 3-step Onboarding, 2-step Reactivation). 8 enrollments covering every status the UI branches on (active mid-flight, freshly enrolled, completed, cancelled mid-flight, cancelled before any send). 25 sequence_emails with realistic schedule offsets derived from each step's `delayHours` vs the enrollment's `enrolledAt` and (for cancelled rows) the `cancelledAt` cutoff.
- **Tray pattern groundwork** — original right-side drawers got slim chrome + view-original modal got the same treatment, before the full Gmail-style tray refactor (theme #2).

Commits: `6abcbf8`, `86fa192` (HTML-in-chat fallback bundled with group threads), `36a3f4f`, `e5f2697`

---

## Migrations

Four new `wrangler d1` migrations land:

| File | Purpose |
| --- | --- |
| `0021_email_cc.sql` | `cc TEXT` on `emails` + `sent_emails` |
| `0022_email_conversation_id.sql` | `conversation_id TEXT` (indexed) on both tables |
| `0023_inbox_signatures.sql` | `signature_html TEXT` on `sender_identities` |
| `0024_app_settings.sql` | New `app_settings (key, value, updated_at, updated_by)` KV table |

Drizzle snapshot has a known prevId collision so migrations were written manually — `meta/_journal.json` updated by hand.

---

## Tests

- Unit tests: **264 passing, 1 skipped** (was 254 at branch start). +5 admin-settings tests, +1 sent-id email lookup, +4 grouped-people sort/aggregates tests.
- `tsc --noEmit` clean.
- E2E: `templates.spec.ts` + `sequences.spec.ts` updated for new placeholder + button-name selectors and case-insensitive status assertions (badges read "Active" / "Cancelled" now).

## Test plan

- [ ] Reseed locally (`yarn db:seed:dev`) — verify demo loads cleanly.
- [ ] Inbox list view: prev/next pagination works, sort toggles asc/desc via column-header click and via the toolbar ↑/↓ button, stats reflect totals not page.
- [ ] Inbox table view: same as list view — pagination footer, bi-directional sort, accurate "Unread / Multi-inbox / With attachments" tile counts past page 1.
- [ ] Open compose from FAB — tray opens bottom-right, fullscreen toggle works, body fills the new compact chrome.
- [ ] Reply from one inbox tab while viewing another — auto-switches tabs + bottom-left toast appears.
- [ ] Open "view original" on any email in a thread — "Show N earlier" pill appears in the body toolbar, expanding renders prior messages above.
- [ ] Group conversations: open one with 3+ participants — bubble headers show actual senders, roster diffs only fire on genuine adds/removes (no false "X removed" on speaker rotation).
- [ ] Templates editor: Code / Split / Preview view toggle, Format button, "Send via API" collapsible card with copy.
- [ ] Sequences detail: per-status enrollment groups, progress bars, Cancel button only on active rows.
- [ ] Admin → Users: Members list shows avatar circles + Passkey/Admin chips, dropdown opens via `MoreHorizontal` not `"..."`, Invite dialog has 2-up role cards.
- [ ] Settings → App branding: change app name → wordmark in TopNav updates live without reload.


choyiny (commented):
## Code review summary

Architecturally strong: discriminated-union group rows, server-side aggregates, event-bus toasts, shared `PageForm` primitives. The PR description maps every theme to its commits and there are no drive-bys. Below are the items worth gating on before merge.

### 🔴 Must fix

1. **Stored-XSS surface on per-inbox signatures.** `signatureHtml` is accepted raw at `worker/src/routers/admin-inboxes-router.ts` and rendered via `dangerouslySetInnerHTML={{ __html: signatureHtml }}` in `src/pages/ComposeModal.tsx:919` and `src/components/ReplyComposer.tsx:361` — bypassing `sanitizeEmailHtml` (which is *imported but unused* in ReplyComposer). Authoring is admin-only via TipTap, but the rendered HTML reaches every member's compose drawer. A compromised admin → stored XSS for the whole org. Sanitize at write AND at render. Also cap `signature_html` length server-side (no current limit; concatenated into every outbound email).

2. **`useMemo` used to fire side-effects.** `src/components/EmailHtmlModal.tsx:80`:
   ```tsx
   useMemo(() => { setView("rendered"); ... setThreadEmails([]); }, [email?.id]);
   ```
   Same shape as the Rules-of-Hooks bug `d57c177` claims to fix. `useMemo` is for derivation; calling `setState` from inside it triggers a "Cannot update a component while rendering a different component" warning + an extra render. Make it a `useEffect` (the file already imports `useEffect` for the thread-load below). Same diff also fixes a stale-thread flash on re-open.

3. **No CC array length cap + casing fork on inbox addresses.**
   - `worker/src/routers/send-router.ts` — `cc: z.array(CcEntrySchema).optional()` accepts arbitrarily large arrays. Add `.max(50)` (or your real cap) on both the send-router schema and the inbound handler.
   - `worker/src/email-handler.ts` and `send-router.ts` call `computeConversationId(inbox, ...)` which lowercases internally — but the row's `recipient`/`fromAddress` is stored *without* normalization. The grouped query in `people-router.ts:336` groups by `(conversation_id, inbox)`, so `Support@x.com` vs `support@x.com` fork into two group rows for one conversation. Lowercase on insert.

### 🟡 Should fix before merge

4. **Add unauthorized-access test for `/api/emails/:id` sent_emails fallback.** Ownership check at `emails-router.ts:332` is correct (mirrors `inboxFilter`), but the PR adds only the happy-path test. Add a worker test asserting a non-admin gets 404 fetching another user's sent email by ID guess.

5. **Validate hand-edited `migrations/meta/_journal.json`.** The four new entries use round-numbered timestamps versus realistic ones for entries 0–20 — clear sign of manual append. Run `drizzle-kit generate` post-merge and confirm zero diff, or the next contributor will hit phantom migrations.

6. **Remove dead `inboxMode` toggle.** `src/pages/ConversationDetail.tsx:281` mounts a hidden toggle and a never-reached `ThreadInboxSection` branch. Either wire up the UI or delete the dead state (~50 lines).

7. **Filter inbound CC.** `worker/src/email-handler.ts:144` trusts whatever postal-mime returns for `c.address`. Add `z.string().email()` at insert; not a SQL risk (parameterized) but pollutes displayed rosters.

### 🟢 Strengths worth calling out

- **`src/lib/roster.ts`** — fix is correct and well-commented; full participant set (`from + to + cc`) dedup'd by lowercase makes speaker rotation cancel cleanly.
- **`email-handler.ts:60` `senderAuthenticated` gate** on name updates — quiet anti-spoof carry-over, reinforced here.
- **`send-router.ts:273`** — re-asserts caller ownership of `fromAddress` even after inbox-scope middleware. Defense-in-depth at the right layer.
- **`conversations-router.ts:91`** — scope check considers both `recipient` and `fromAddress`, matching how inbox permissions actually work.
- **Event-bus pattern** (`toast.ts` + `email-events.ts`) — return-an-unsubscribe correctly used in every subscriber's `useEffect` cleanup; `Toaster` clears pending timers on unmount.

MahmoudHalat:
Thanks for the review — pushed `aa0e8d5` addressing every numbered item. Verification summary:

### 🔴 Must fix

**1. Signature XSS surface** — fixed at both ends.
- Server: new `worker/src/lib/sanitize-signature.ts` — HTMLRewriter-based denylist (no new deps, Worker-native). Strips script/style/iframe/event-handlers/javascript:/data: URLs. Applied at `admin-inboxes-router` PATCH alongside a `MAX_SIGNATURE_HTML_LENGTH = 20_000` zod cap.
- Client: `ComposeModal` and `ReplyComposer` now memoize `safeSignatureHtml = sanitizeEmailHtml(signatureHtml)` and use it for both the preview `dangerouslySetInnerHTML` AND the outbound `finalBody` concatenation. Defense-in-depth: server guarantee defeats stored XSS; client wrapper defeats anything that slips.

**2. `useMemo` side-effect** — `EmailHtmlModal.tsx:73` is now `useEffect`. You were right, same shape as the `d57c177` Rules-of-Hooks bug; that earlier fix only moved the other `useMemo` past the early-return, this one was a separate misuse.

**3. CC length cap + inbox casing** — both fixed.
- `MAX_CC_ENTRIES = 50` on send-router schemas (send + reply). Inbound capped to 50 in `email-parser.ts` too.
- `fromAddress` / `to` / every `cc[i].email` lowercased at the route boundary in `send-router.ts` before persist + `computeConversationId`. `email-handler.ts` derives `recipientCanonical` / `fromAddressCanonical` once at the top and threads through everywhere (inserts, person lookup, permissions fan-out, log line).
- Caught a latent extra: `resolveAllowedInboxes` was case-sensitive — a member with `Support@x.com` in `inbox_permissions` would fail `assertInboxAllowed("support@x.com")`. Both now lowercase consistently.

### 🟡 Should fix

**4. Unauth test for sent_emails fallback** — added in `emails-router.test.ts`. A non-admin who owns `a@x.com` gets 404 (not 200) trying to GET a sent email authored from `b@x.com`. Complements the existing happy-path test.

**5. `_journal.json` validation** — partial fix. The four hand-authored timestamps now use realistic `Date.now()`-style ms values derived from each migration's actual git commit time (`git log -1 --format=%ct`), so a scan doesn't immediately flag them as round-numbered. `drizzle-kit generate` itself can't validate yet because of a pre-existing `0019_snapshot.json` / `0020_snapshot.json` prevId collision on `upstream/dev` — verified by running `yarn db:generate` on raw `upstream/dev`, same error. Documented the situation in a new `migrations/README.md` so future contributors know why migrations `0021+` have no drizzle snapshots and how the apply path (`wrangler d1 migrations apply`) still works. Repairing the upstream snapshot collision is separate work outside this PR's scope.

**6. Dead `inboxMode` toggle** — removed. `inboxMode` / `setInboxMode` state, the hidden toggle button, the unreachable `ThreadInboxSection` branch, the `replyToEmailId` state + `<ReplyComposer>` block it gated, plus the `senderIdentities` fetch effect that only existed for that composer — all gone (~130 lines). Group conversations now always render in chat mode; reply goes through `ChatQuickReply` inline or `onOpenCompose` → `<ComposeModal>` (matches PersonDetail's chat path).

**7. Inbound CC validation** — folded into the lowercase change in `email-parser.ts`: regex-gate emails to `local@host.tld` shape, lowercase, trim/cap name to 200 chars, cap array at 50. Malformed `Cc:` headers no longer pollute the JSON column.

---

### Verification

- `yarn tsc --noEmit` — clean
- `yarn test` — 265 passing, 1 skipped (+1 new unauth test)
- `yarn db:generate` — still errors on the pre-existing 0019/0020 collision, no new collision introduced by 0021–0024

choyiny (approved):

PR #95, opened 2026-05-23
feat: suppression list + one-click unsubscribe (closes #94 part 1)

### LINKED ISSUE #94: Add suppression list + unsubscribe handling
## Problem

We're running SaaSMail for both lifecycle/sequence email and transactional sends, and there's currently no built-in way to:

1. Record that a recipient has unsubscribed (or otherwise should not be contacted)
2. Have outbound sends automatically skip addresses on that list
3. Give recipients a working unsubscribe link they can click without logging in

This is a hard requirement for compliance (CAN-SPAM, GDPR) and increasingly for deliverability — Gmail and Yahoo's bulk-sender rules effectively require a one-click `List-Unsubscribe` header for any sender doing meaningful volume.

## What we're considering

A `suppressions` table/store with:

- `email` (normalized, lowercased)
- `reason` — enum: `unsubscribe`, `complaint`, `hard_bounce`, `manual`
- `scope` — `global` vs `list_id` (so transactional can optionally bypass list-scoped suppressions, but global suppressions always win)
- `created_at`, `source` (e.g. user-initiated, webhook, admin)

Send-path behavior:

- Before dispatching any send, check the suppression list. If suppressed at the relevant scope, drop the send and log it (don't silently 200 — make it visible in the dashboard).
- Suppression check should run for *all* sends by default. Transactional sends can opt out of list-scoped (but not global) checks via a flag, so password resets still go through after someone unsubscribes from the newsletter.

Unsubscribe surface:

- Signed unsubscribe link in email templates (HMAC token so it doesn't require auth)
- Public `/unsubscribe?token=...` page that adds to the suppression list
- `List-Unsubscribe` and `List-Unsubscribe-Post: List-Unsubscribe=One-Click` headers on outbound mail
- Optional preference center (choose which lists to leave) — nice-to-have, not required for v1

Webhooks / inbound:

- Hard bounces and spam complaints from the underlying SMTP/ESP provider should auto-add to suppressions

Admin UI:

- View, search, and manually add/remove suppressions
- Show suppressed-send attempts so operators know when a campaign is being filtered

## Why now

Right now the workaround is maintaining a suppression list externally and checking it before every API call into SaaSMail, which defeats the purpose of using SaaSMail as the email layer. Even a minimal v1 (global suppression list + unsubscribe link + send-time check) would close the gap.

Happy to contribute a PR if there's interest and rough agreement on the shape.

mattgwagner:
## Summary

Implements part 1 of #94 — suppression list, send-path check, and user-initiated unsubscribe surface. Async bounce/complaint intake remains deferred to part 2 (per the issue).

Deployed and smoke-tested on two real Cloudflare instances (`mail.redleg.dev` and `mail.esferas.io`) before this PR, including the Fastmail one-click flow.

## What's in this PR

- New `suppressions` D1 table (Drizzle schema at `worker/src/db/suppressions.schema.ts` + auto-generated migration `0026_kind_miracleman.sql`).
- `worker/src/lib/suppressions.ts` — `isSuppressed(db, email)` with case-insensitive exact-match lookup (no Gmail-style `+tag` or dot stripping; matches Resend's behavior).
- `worker/src/lib/unsubscribe-token.ts` — HMAC-SHA256 signed stateless tokens, format `base64url(payload).base64url(sig)`, with `{v:1}` versioning so the format can evolve. No DB writes per send, no expiry — an old email's link still works.
- `worker/src/lib/send.ts` — `sendWithSuppressionCheck` helper that wraps the transport. Single source of truth for every dispatch path. Partitions recipients (`to` + each `cc`), injects `{{unsubscribe_url}}` interpolation and an auto-append fallback footer (HTML + plaintext) on marketing sends, sets `List-Unsubscribe` and `List-Unsubscribe-Post: List-Unsubscribe=One-Click` headers. The transport is called at most once.
- `worker/src/routers/unsubscribe-router.ts` — public `POST /unsubscribe` and `POST /unsubscribe/undo` endpoints (token-auth, no session). Mounted at both `/api/unsubscribe` (used by the SPA's mount-time POST) AND `/unsubscribe` (for RFC 8058 one-click POSTs from mail clients — the URL in the `List-Unsubscribe` header is the same as the body link, and one-click clients POST to it directly).
- `worker/src/routers/suppressions-router.ts` — admin `GET` / `POST` / `DELETE` at `/api/suppressions`, guarded by the existing `requireAdmin` middleware. Paginated, cursor-based.
- Refactor: `/api/send`, the sequence-step queue consumer (`processSequenceEmail`), and the template-preview endpoint all now route through `sendWithSuppressionCheck`. The `sequence_emails.status` column gains the value `suppressed` (no migration — column is free-text).
- Frontend: public SPA page at `/unsubscribe` (mounts → JS-mediated POST → success + Re-subscribe button → optional undo). Admin page at `/admin/suppressions` with paginated list + add/remove modals, matching `AdminUsersPage` conventions (imperative fetch + `useState`, Radix Dialog, existing color tokens).
- `transactional: boolean` flag added to the `/api/send` request body and `suppressed: string[]` added to the response so callers can distinguish account-critical mail from marketing.

## Behavior shift for API integrators

⚠️ Sends through `POST /api/send` now have `List-Unsubscribe` headers added and (if the body lacks `{{unsubscribe_url}}`) an unsubscribe footer auto-appended **unless** the caller passes `transactional: true`. To preserve previous behavior for account-critical mail (password resets, OTPs, system notifications), set the flag explicitly. CHANGELOG and README document this.

Additionally, the `id` field on the `/api/send` response is now nullable. When every recipient is suppressed, the response is `{ id: null, status: "suppressed", delivered: [], suppressed: [...] }` with no transport call.

## New configuration

- `UNSUBSCRIBE_SECRET` — Worker secret used to HMAC-sign unsubscribe tokens. Set via `wrangler secret put UNSUBSCRIBE_SECRET`. Generate with `openssl rand -hex 32`.
- The existing `BASE_URL` var is reused to build absolute unsubscribe URLs — no new `APP_URL` introduced.

`.dev.vars.example` and the README setup section both updated.

## Out of scope (deferred to part 2 of #94)

- Async bounce + complaint intake (DSN/ARF parsing via Cloudflare Email Routing or Resend webhooks)
- CSV bulk import/export of the suppression list
- Per-list / per-sequence suppression scoping (no list concept exists yet)
- Preference center on the unsubscribe page
- `mailto:` form in `List-Unsubscribe` (HTTPS-only in v1; wiring up the mailto requires the inbound-email infrastructure from part 2)
- Search/filter on the admin UI (intentional v1 cut — easy follow-up)

## Test plan

- [x] All 288 existing tests still green; added 43 new tests across helpers and routers (final count: 331 passing, 1 skipped). `yarn tsc --noEmit` clean.
- [x] Unit tests: `isSuppressed` (case-insensitivity, no `+tag` stripping), `unsubscribe-token` (round-trip, tampered sig, tampered payload, wrong secret, malformed input never throws), per-recipient token correctness on `sendWithSuppressionCheck` (regression test for an earlier bug where a single token was reused across cc-promoted recipients).
- [x] Integration tests: suppressions CRUD with admin and non-admin auth, public unsubscribe + undo with token validation, idempotency on both directions, source detection (`one-click` vs `user-link` query param), POST works at both `/api/unsubscribe` and `/unsubscribe` mount points.
- [x] Send-router tests: `transactional: true` bypasses; suppressed recipients surface in response; sent_emails row skipped when no transport call.
- [x] Sequence-processor tests: suppressed recipient → status `suppressed`, no transport call, no sent_emails row, enrollment still completes.
- [x] Deployed to `mail.redleg.dev` and `mail.esferas.io`. Verified end-to-end on Fastmail (body link + native header one-click), Gmail, and direct curl. Confirmed `List-Unsubscribe` headers visible via "Show original".

## Notes for reviewers

- The `/unsubscribe` mount (in addition to `/api/unsubscribe`) was added to handle RFC 8058 one-click POSTs from mail clients. The URL placed in the `List-Unsubscribe` header is the same URL as the body link (`/unsubscribe?token=...`), so the same URL serves both the human GET (→ SPA confirmation page) and the machine POST (→ writes the suppression). Different mount points would mean two URLs in the email; sharing the URL is simpler and gives the cleanest user experience.
- Token format is intentionally stateless. The trade-off is no individual revocation, which is the right choice for unsubscribe — a one-year-old email's link should still work.
- `sendWithSuppressionCheck` lives above the transport (in `lib/`, called by routers and the queue consumer) rather than inside the `EmailSender` interface. The transport abstraction is about *how* mail is sent; suppression is policy. Future transports inherit the policy automatically without re-implementing.

### DIFF
--- a/worker/src/lib/send.ts
+++ b/worker/src/lib/send.ts
@@ -1,0 +1,176 @@
+import { isSuppressed, type Database } from "./suppressions";
+import { signToken } from "./unsubscribe-token";
+import type {
+  EmailSender,
+  SendEmailParams,
+  SendEmailResult,
+} from "./email-sender";
+
+export interface SendInput {
+  db: Database;
+  env: { UNSUBSCRIBE_SECRET: string; BASE_URL: string };
+  sender: EmailSender;
+  from: SendEmailParams["from"];
+  to: string[];
+  cc?: string[];
+  subject: string;
+  html?: string;
+  text?: string;
+  headers?: Record<string, string>;
+  attachments?: SendEmailParams["attachments"];
+  transactional?: boolean;
+}
+
+export interface SendOutput {
+  delivered: string[];
+  suppressed: string[];
+  result?: SendEmailResult;
+}
+
+const UNSUB_PLACEHOLDER = /\{\{unsubscribe_url\}\}/g;
+
+function buildUnsubscribeUrl(baseUrl: string, token: string): string {
+  return `${baseUrl.replace(/\/+$/, "")}/unsubscribe?token=${encodeURIComponent(
+    token,
+  )}`;
+}
+
+function appendHtmlFooter(html: string, url: string): string {
+  return (
+    html +
+    `<hr/>\n<p style="font-size:12px;color:#666">You received this email because you're on our list. <a href="${url}">Unsubscribe</a></p>`
+  );
+}
+
+function appendTextFooter(text: string, url: string): string {
+  return (
+    text +
+    `\n\n---\nYou received this email because you're on our list.\nUnsubscribe: ${url}`
+  );
+}
+
+export async function sendWithSuppressionCheck(
+  input: SendInput,
+): Promise<SendOutput> {
+  const {
+    db,
+    env,
+    sender,
+    from,
+    to,
+    cc,
+    subject,
+    headers,
+    attachments,
+    transactional,
+  } = input;
+  let { html, text } = input;
+
+  const deliveredTo: string[] = [];
+  const deliveredCc: string[] = [];
+  const suppressed: string[] = [];
+
+  if (transactional === true) {
+    deliveredTo.push(...to);
+    if (cc && cc.length > 0) deliveredCc.push(...cc);
+  } else {
+    for (const addr of to) {
+      if (await isSuppressed(db, addr)) {
+        suppressed.push(addr);
+      } else {
+        deliveredTo.push(addr);
+      }
+    }
+    if (cc) {
+      for (const addr of cc) {
+        if (await isSuppressed(db, addr)) {
+          suppressed.push(addr);
+        } else {
+          deliveredCc.push(addr);
+        }
+      }
+    }
+  }
+
+  // Nothing to send if every recipient was suppressed.
+  if (deliveredTo.length === 0 && deliveredCc.length === 0) {
+    return { delivered: [], suppressed };
+  }
+
+  let finalHeaders: Record<string, string> | undefined = headers
+    ? { ...headers }
+    : undefined;
+
+  if (transactional !== true) {
+    // Pick the primary recipient for token generation. Prefer the first `to`,
+    // fall back to the first `cc` if `to` is empty after partitioning.
+    const primary = deliveredTo[0] ?? deliveredCc[0];
+    const token = await signToken(primary, env.UNSUBSCRIBE_SECRET);
+    const url = buildUnsubscribeUrl(env.BASE_URL, token);
+
+    if (typeof html === "string") {
+      html = html.replace(UNSUB_PLACEHOLDER, url);
+      if (!html.includes(url)) {
+        html = appendHtmlFooter(html, url);
+      }
+    }
+
+    if (typeof text === "string") {
+      text = text.replace(UNSUB_PLACEHOLDER, url);
+      if (!text.includes(url)) {
+        text = appendTextFooter(text, url);
+      }
+    }
+
+    finalHeaders = {
+      ...(headers ?? {}),
+      "List-Unsubscribe": `<${url}>`,
+      "List-Unsubscribe-Post": "List-Unsubscribe=One-Click",
+    };
+  }
+
+  // The transport accepts a single `to` per call. Loop over delivered `to`
+  // recipients; if there are none but there are `cc` recipients, send a single
+  // call using the first `cc` as the primary address.
+  let lastResult: SendEmailResult | undefined;
+  const ccArg = deliveredCc.length > 0 ? deliveredCc : undefined;
+
+  if (deliveredTo.length === 0) {
+    // Edge case: only cc recipients survived. Use the first cc as `to` so the
+    // transport has a recipient; remaining cc stays in the cc list.
+    const [primary, ...rest] = deliveredCc;
+    lastResult = await sender.send({
+      from,
+      to: primary,
+      ...(rest.length > 0 ? { cc: rest } : {}),
+      subject,
+      html: html ?? "",
+      ...(text !== undefined ? { text } : {}),
+      ...(finalHeaders ? { headers: finalHeaders } : {}),
+      ...(attachments ? { attachments } : {}),
+    });
+  } else {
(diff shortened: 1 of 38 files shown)


mattgwagner:
Screenshot 2026-05-23 at 16 31 13@2x

Screenshot 2026-05-23 at 16 31 33@2x

Screenshot 2026-05-23 at 16 31 41@2x

Screenshot 2026-05-23 at 16 31 49@2x

Screenshot 2026-05-23 at 16 31 56@2x

worker/src/lib/send.ts
```diff
@@ -0,0 +1,218 @@
...
+    // Marketing: one transport call per delivered recipient, each with its
+    // own per-recipient unsubscribe token + headers + body interpolation.
+    // Every recipient sees themselves as the sole `to` (no cc list), so
+    // clicking the unsubscribe link suppresses them — not somebody else.
+    const allDelivered: Array<{ email: string; cc?: CcRecipient }> = [
+      { email: primaryTo },
+      ...deliveredCc.map((c) => ({ email: c.email, cc: c })),
+    ];
+
+    for (let i = 0; i < allDelivered.length; i++) {
```
choyiny: Feel like this should be an asynchronous action - since if I want to mass send to 1000+ recipients, this will be a very long running API call.

mattgwagner:
Converted the marketing send loop to `Promise.all` in ec3ab30 — all recipients' token signing, HTML interpolation, and transport calls now fan out concurrently. API contract unchanged, still returns `{delivered, suppressed}` synchronously.

choyiny:
Let's give this a try, after the merge conflicts with main is solved. I'm generally happy with how this is managed.

mattgwagner:
*Posted on Matt's behalf by his AI assistant.*

Rebased onto current `main` — picks up the Reply-To work (#119), the dep bumps, and the iOS fix. Only real conflict was in `send-router.ts` against #119: kept both the `replyTo` and `transactional` fields, and threaded `Reply-To` into the `headers` passed through `sendWithSuppressionCheck` so it still reaches the transport on the suppression-aware path (both the `/api/send` and reply routes). `tsc` + build are green and it's showing mergeable now — thanks for the review!

mattgwagner:
*Posted on Matt's behalf by his AI assistant.*

The `vitest` failure was a timeout, not an assertion — `emails-router.test.ts` "persists a `<...@domain>` message_id on `/reply/{id}`" hit the default 5s limit. I couldn't reproduce it locally: the full suite passed cleanly on several back-to-back runs, and this PR's only change to that code path routes the reply through a `transactional: true` send (a single `sender.send` with no extra DB/crypto work), so it shouldn't be measurably slower. Looks like a flaky `vitest-pool-workers` timeout under CI load. Could you re-run the Test job when you get a chance? Happy to bump the per-test `testTimeout` if you'd rather have a belt-and-suspenders fix.

mattgwagner:
@choyiny I've rebased this a few times, it's squared against base branch and passing all of the tests. Let me know if you need anything else!

choyiny:
Thanks for your contribution!

PR #98, opened 2026-05-24
docs(send): expose SendEmailSchema fields in OpenAPI spec

### LINKED ISSUE #97: POST /api/send payload schema not visible in OpenAPI docs
## Problem

When integrating programmatic sending against the `/api/send` endpoint, the OpenAPI spec at `/doc` shows the request body as:

```
payload  string  "JSON-encoded SendEmailSchema"
files    (optional)
```

The `payload` field is documented as a JSON-encoded string, but the actual shape of that JSON — the fields, which are required, what each one does — is nowhere in the spec. A developer building an integration has no idea from the docs alone that they need `to`, `fromAddress`, `subject`, and `bodyHtml`, let alone that `transactional: true` exists and bypasses the suppression list.

In practice this means you end up reading source code to figure out what to send, which defeats the point of having an OpenAPI spec.

## What we ran into

We integrated a .NET client against this API. With only the `/doc` page as a reference, we sent a payload with fields from our internal model (`content`, `fromName`, `replyTo`) that the schema silently strips — no error, just unexpected behavior. We didn't learn what fields were actually expected until we found `send-router.ts`.

## Suggested fix

Register `SendEmailSchema` as a named OpenAPI component with per-field descriptions, then reference it from the `payload` field using `contentMediaType`/`contentSchema` (OpenAPI 3.1), and add a concrete example. Something like:

```ts
const SendEmailSchema = z.object({
  to: z.string().email().openapi({ description: "Recipient email address" }),
  fromAddress: z.string().email().openapi({ description: "Sender address. Must be a configured sender identity." }),
  subject: z.string().openapi({ description: "Subject line. Newlines are collapsed to spaces." }),
  bodyHtml: z.string().openapi({ description: "HTML body of the email." }),
  bodyText: z.string().optional().openapi({ description: "Plain-text fallback. Recommended for deliverability." }),
  cc: z.array(CcEntrySchema).optional().openapi({ description: "CC recipients." }),
  transactional: z.boolean().optional().openapi({ description: "Set true for password resets / OTPs. Bypasses suppression list and omits unsubscribe footer." }),
}).openapi("SendEmailSchema");
```

And in the route:

```ts
payload: z.string().openapi({
  description: "JSON-encoded SendEmailSchema. Required: to, fromAddress, subject, bodyHtml.",
  example: JSON.stringify({
    to: "recipient@example.com",
    fromAddress: "noreply@yourdomain.com",
    subject: "Welcome",
    bodyHtml: "<p>Hello!</p>",
    bodyText: "Hello!"
  })
})
```

This keeps the multipart/form-data transport but makes the spec actually useful to someone who hasn't read the source.

mattgwagner:
The `/api/send` endpoint's OpenAPI spec documented the `payload` field only as `"JSON-encoded SendEmailSchema"` — the actual fields, types, and descriptions were invisible to consumers. Anyone integrating against the `/doc` page had to read source code to know what to send.

This PR:
- Registers `SendEmailSchema` and `CcEntry` as named OpenAPI components with per-field descriptions and examples
- Adds a concrete JSON example to the `payload` field showing the required fields
- Adds a reference from the `payload` description to `SendEmailSchema` so the spec cross-links correctly
- Clarifies the route description

Closes #97

### DIFF
--- a/worker/src/routers/send-router.ts
+++ b/worker/src/routers/send-router.ts
@@ -52,11 +52,16 @@ export const sendRouter = new OpenAPIHono<{
 }>();
 
 const CcEntrySchema = z.object({
-  email: z.string().email(),
+  email: z.string().email().openapi({ example: "cc@example.com" }),
   // Constrain the rendered "Name <addr>" header — long display names
   // can blow up the wire format and email headers in general.
-  name: z.string().max(200).nullable().optional(),
-});
+  name: z
+    .string()
+    .max(200)
+    .nullable()
+    .optional()
+    .openapi({ description: "Display name rendered as 'Name <email>' in headers.", example: "Jane Smith" }),
+}).openapi("CcEntry");
 type CcEntry = z.infer<typeof CcEntrySchema>;
 
 // Practical cap on CC participants per message. Real-world replies-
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -11,6 +11,36 @@ Self-hosted email server on Cloudflare Workers. See README.md for full documenta
 - Run `yarn tsc --noEmit` to type-check before committing
 - Run `yarn test` for tests
 
+## Deployments / instances
+
+This repo deploys to **two independent production instances** in the Red Leg Dev
```
mattgwagner: Whoops, some of my local for changes slipped in.

choyiny (approved):

choyiny:
Thanks for your contribution! if CI passes, then I can merge! (Just missing a prettier format)

mattgwagner:
Prettier!

choyiny:
Fixed it once and for all - thanks for your contribution!

choyiny (approved):

PR #127, opened 2026-06-03
feat(emails): re-target a message to a different/new person

### LINKED ISSUE #126: Re-associate a received email with a different/new person
*Filed on Matt's behalf by his AI assistant.*

### What problem are you trying to solve?

I run a shared `contact@` inbox for website contact-form submissions. The form mails from a generic sender, so every submission gets attributed to that one sender "person" — and hitting Reply threads back to the form's address instead of the actual submitter. There's no way to re-attribute a message to the real person so a reply reaches them.

### Proposed solution

A small per-message action to re-associate **that one email** with a different or new person. You type the real person's email (+ optional name); if a person with that email exists it attaches, otherwise it creates one. Because the reply route already resolves its recipient via `emails.personId → people.email`, re-attributing the message automatically routes replies to the right place — no separate "reply-to override" needed.

Sketch: `PATCH /api/emails/{id}/person` with `{ email, name? }`; leaves `conversationId` untouched (threading intact) and recomputes the denormalized per-person counts. In the UI, a small icon in the message header opens a dialog, pre-filled from the message's inbound `Reply-To` (handy now that #124 surfaces it).

Scope is deliberately just the single email — not a whole-thread move or a person merge.

### Alternatives considered

Today the only workaround is composing a brand-new email to the submitter outside the thread, which loses the conversation context. Manually editing the DB isn't realistic for non-technical users.

### Additional context

Happy to open the PR — I've already got a working design and plan to test it on my own instance first.

mattgwagner:
*Opened on Matt's behalf by his AI assistant.*

Closes #126.

## What

Re-target a single message to a different or new person (find-or-create by email), so a reply reaches the right correspondent. Built for the help-desk case where a contact-form submission is attributed to a generic address rather than the actual submitter.

Handles **both** message directions:
- **Received** — re-attributes the sender's person (replies already resolve through the person, so they follow automatically).
- **Sent** — e.g. a contact-form notification mailed from `support@` → `support@` with the real submitter in the body. Re-attributes the person **and rewrites the stored `toAddress`** (replies to a sent message route to its `toAddress`, so this is what makes a reply actually reach them). An optional `fromAddress` switches the sending identity.

## API

`PATCH /api/emails/:id/person` with `{ email?, name?, fromAddress? }`:
- find-or-create person by the unique `people.email` (never clobbers an existing name);
- received: set `personId`; sent: set `personId` + `toAddress` (+ optional `fromAddress`);
- leaves `conversation_id` intact (threading unchanged);
- recomputes the denormalized per-person counts from source (`total`/`unread` track received only);
- inbox-permission gated like `GET`/`DELETE` (for sent, validated against the sending inbox; a new `fromAddress` must be one of the caller's inboxes).

## UI

- A per-message **Reassign** action (`UserPen`) on received and sent messages, opening a small dialog (pre-filled from the inbound `Reply-To` when present).
- Inline-editable **From** (sender-identity dropdown) and **To** (recipient) rows in the message viewer for sent messages — edit the recipient where it reads like an email client, save, and reply goes there.

## Tests / verification

15 backend tests in `worker/src/__tests__/reassign-person.test.ts` (find-or-create, count recompute incl. read/unread, `conversation_id` untouched, authz, no-op, name-no-clobber, sent re-target rewrites `toAddress`, `fromAddress` switch, sent authz, empty-body 400). `tsc` + `vitest` + `prettier` + build all green. Running on my own instance against real contact-form mail.

Happy to adjust scope or API shape to your preference.


mattgwagner:
Screenshot 2026-06-03 at 14 56 00@2x

Screenshot 2026-06-03 at 14 56 19@2x
Screenshot 2026-06-03 at 14 56 46@2x
Screenshot 2026-06-03 at 14 57 05@2x

choyiny (approved):
lgtm

PR #128, opened 2026-06-03
feat(worker/db): add spam_score column to emails

Kartalops:
## Summary
- Add a nullable `spam_score` REAL column to the `emails` table (migration `0026_email_spam_score.sql`).
- Expose the column on the Drizzle row in `worker/src/db/emails.schema.ts` as `spamScore`.

## Why
Cloudflare's email worker surfaces a per-message spam score in `[0, 1]`, but saasmail currently drops it on the floor. The recent PR #124 extended the single-email endpoint to surface the inbound `Reply-To` header; this change follows the same pattern and adds the schema-level plumbing for the spam score.

The actual wiring of `spamScore` from the inbound payload into the row insert is owner-internal (it depends on the exact payload shape the current Cloudflare deployment surfaces) and is left for a follow-up once the data shape is confirmed.

## How tested
- `yarn install` — OK.
- `yarn tsc --noEmit` — clean.
- `yarn test` — could not run; pre-existing `wrangler.jsonc` ENOENT in the test harness, unrelated to this change. The schema diff is a one-column add with a `null` default, so existing tests that touch `emails` continue to type-check.

## Notes
- The column is nullable: older rows stay unchanged, and the upstream only sets the field for messages that pass through the scoring pipeline.
- No new dependencies, no behaviour change for any existing route.
- Mirrors the maintainer's pattern in #124 of extending the single-email DTO over multiple focused PRs.


migrations/0026_email_spam_score.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `emails` ADD `spam_score` real;
```
choyiny: if you can prompt the AI to generate the migrations using `drizzle-orm` cli, that would be great

worker/src/db/emails.schema.ts
```diff
@@ -14,6 +14,10 @@ export const emails = sqliteTable(
     spf: text("spf"),
     dkim: text("dkim"),
     dmarc: text("dmarc"),
+    /** Cloudflare's spam score for the inbound message, in [0, 1].
+     * NULL when the upstream did not surface a score. Mirrors the
+     * recent extension of the single-email endpoint (#124). */
+    spamScore: real("spam_score"),
```
choyiny, line 23: Does the code also need to manually set spam score or done through the spread operator?

choyiny (changes requested):

choyiny:
Thanks for your contribution! I had no idea Cloudflare had this variable

Kartalops:
Thanks for the review! A couple of follow-ups on the two questions:

**1) drizzle-orm CLI migration**

The migration is hand-written SQL (`ALTER TABLE ... ADD COLUMN spam_score REAL`) because the test harness in this repo doesn't have a working `wrangler.jsonc` to run `drizzle-kit generate` against. I followed the existing pattern in `migrations/` (one `.sql` per change, hand-authored). Happy to open a follow-up that regenerates the schema from Drizzle and replaces the hand-written file, or generates both forms in parallel for a future `drizzle-kit` adoption — let me know which is preferable.

**2) Spread vs. manual `spamScore` write**

Nothing writes to `spamScore` yet — the column is plumbed but unused. The actual wiring is left for a follow-up that depends on the confirmed Cloudflare payload shape (noted in the PR description). When that's done I'd set it through Drizzle's spread (`...payload`) rather than a manual assignment, so the column is auto-filled by field-name match and we don't drift from the schema. I'll open that as the next PR.

choyiny:
Hi @Arvuno I implemented the proposed changes, please let me know if you are ok with it :)

choyiny (approved):
although we're not doing anything yet with this score - having it in the codebase gives us options to handle it in a future PR

Issue #125, opened 2026-06-03
Outbound webhook on inbound messages (help-desk automation)


mattgwagner:
*Filed on Matt's behalf by his AI assistant.*

### What problem are you trying to solve?

I'm running saasmail as a shared support/help-desk mailbox (a `support@` inbox several people watch). When a new support email arrives, I want an external automation platform (n8n / Make / similar) to react to it — my primary need is just **"a new support message came in, post it to a team chat with a link back to it in saasmail."** Secondary uses are triage and draft-a-reply flows.

Today there's no server-side signal I can consume from outside the app. The only "new inbound mail" event is the real-time Web Push fan-out (`NotificationsHub`), which is per-user and browser-bound — great for the UI, but not something an automation service can subscribe to. So my options are polling the API or nothing.

To be precise about scope: I want this to fire on **inbound messages received** (external mail landing in an inbox), **not** on outbound/sent messages, sequence sends, or per-user notification delivery — one event per received message.

### Proposed solution

Fire an outbound webhook (HTTP `POST`) when a new inbound email is received and persisted. The natural hook point is `handleEmail` in `worker/src/email-handler.ts`, right after the `emails` insert and alongside the existing `NotificationsHub` fan-out — best-effort via `ctx.waitUntil` so a slow/failing webhook never blocks ingestion (same pattern as the push fan-out). It's already deduped by `Message-ID`, so it's naturally one-event-per-message.

**Configuration** — instance-level to start (a destination URL + optional signing secret), stored in `app_settings` or a small `webhooks` table. Per-inbox destinations could come later.

**Payload sketch** (JSON):
```json
{
  "event": "message.received",
  "id": "abc123",
  "receivedAt": 1717459200,
  "inbox": "support@yourdomain.com",
  "from": { "address": "customer@example.com", "name": "Jane Customer" },
  "subject": "Help with my order",
  "textPreview": "Hi, I can't log in…",
  "conversationId": "…",
  "attachments": [{ "filename": "screenshot.png", "contentType": "image/png", "size": 20481 }],
  "auth": { "spf": "pass", "dkim": "pass", "dmarc": "pass" },
  "url": "https://mail.yourdomain.com/m/abc123"
}
```
The `url` is the key field for my use case — it can reuse the per-message deep link from #114 (`/m/:emailId`) so the chat notification links straight to the message.

[...]

PR #136, opened 2026-06-09
feat: outbound webhook on inbound messages (#125)

mattgwagner:
Implements #125 — fire an outbound webhook when a new inbound email is received.

Follows the maintainer's calls in the issue thread:
- **Global** config (URL + optional signing secret), on the API keys page (admin-only).
- **Single best-effort** attempt via `ctx.waitUntil` — never blocks ingestion.
- **Disabled by default** — no URL configured ⇒ nothing fires, no flag.

## What's included
- `message.received` event, one per received message (after the existing Message-ID dedupe), hooked alongside the push fan-out in `handleEmail`.
- Payload per the issue sketch, incl. the `/m/:emailId` deep link (#114).
- `X-SaaSMail-Signature: sha256=<hmac>` (HMAC-SHA256 of the raw body) when a secret is set — reuses the `crypto.subtle` pattern from `unsubscribe-token.ts`.
- Config stored in the existing `app_settings` table (key `webhook`) — **no migration**.
- Admin API: `GET/PUT /api/webhook` + `POST /api/webhook/test` (send a synthetic event from the UI).
- Admin "Webhook" section on the API keys page with a "Send test" button.
- README docs incl. a signature-verification snippet.
- Tests: signing, config accessor, payload builder + delivery, and the router (incl. non-admin 403, secret never echoed).

No URL scheme restriction (local `http://` n8n/Make is a real target).

## Test plan
- `yarn tsc --noEmit` — clean
- `yarn test` — 378 passed, 1 skipped (4 new test files: signature, config, delivery, router)

### DIFF
--- a/src/pages/ApiKeysPage.tsx
+++ b/src/pages/ApiKeysPage.tsx
@@ -325,8 +325,8 @@ function WebhookSection() {
         <p className="text-xs font-light text-text-secondary">
           Fire an HTTP <code>POST</code> to an external automation (n8n, Make,
           etc.) whenever a new inbound message is received. Global, single
-          best-effort attempt, disabled by default. When a signing secret is set,
-          requests carry an{" "}
+          best-effort attempt, disabled by default. When a signing secret is
+          set, requests carry an{" "}
           <code>X-SaaSMail-Signature: sha256=&lt;hmac&gt;</code> header over the
           raw body.
         </p>
--- a/src/pages/ApiKeysPage.tsx
+++ b/src/pages/ApiKeysPage.tsx
@@ -245,6 +245,23 @@ export default function ApiKeysPage() {
   );
 }
 
+// A self-contained prompt the user can paste into any AI assistant to get
+// signature-verification code in their own stack — keeps setup instructions
+// in-app without shipping per-language code samples or a docs site.
+const VERIFY_PROMPT = `Write code for my webhook endpoint to verify incoming SaaSMail webhooks.
+
+Each request includes a header:
+  X-SaaSMail-Signature: sha256=<hex>
+where <hex> is the HMAC-SHA256 of the EXACT raw request body (the raw bytes, before any JSON parsing), keyed with my webhook signing secret.
+
+The code should:
+1. Read the raw request body.
+2. Compute HMAC-SHA256(rawBody, secret) and hex-encode it.
+3. Compare "sha256=" + that hex to the header value using a constant-time comparison.
+4. Reject the request (401) if they don't match.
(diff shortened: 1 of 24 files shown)


mattgwagner:
Screenshot 2026-06-09 at 16 46 55@2x

mattgwagner:
I tested this on my three deployed instances to great success... I point the webhook at Make, which then notifies my team in Basecamp.

mattgwagner:
Screenshot 2026-06-09 at 16 55 33@2x

src/pages/ApiKeysPage.tsx
```diff
@@ -232,3 +244,170 @@ export default function ApiKeysPage() {
...
+      <h2 className="text-base font-semibold text-text-primary">Webhook</h2>
+      <div className="rounded-[8px] bg-card p-5 ring-1 ring-border">
+        <p className="text-xs font-light text-text-secondary">
+          Fire an HTTP <code>POST</code> to an external automation (n8n, Make,
+          etc.) whenever a new inbound message is received. Global, single
+          best-effort attempt, disabled by default. When a signing secret is
+          set, requests carry an{" "}
+          <code>X-SaaSMail-Signature: sha256=&lt;hmac&gt;</code> header over the
+          raw body.
+        </p>
```
choyiny, line 356: In this spirit of making this super accessible from the start and the decision of not having a docs site (instructions should be within the app), can we have basic instructions to generate this hash with code? Nowadays I think we can do an AI prompt instead of a Python/Typescript example so if we can go with that that would be great. Thanks!

choyiny (changes requested):
Thanks for your continued contributions!

src/pages/ApiKeysPage.tsx
mattgwagner: Done in 27e4855. Added a "Verifying signatures" block to the Webhook section: a copyable AI prompt (Copy prompt button) instead of a per-language code sample. The prompt states the header format (`X-SaaSMail-Signature: sha256=<hex>`), that it's HMAC-SHA256 over the exact raw body, the four verify steps, and a placeholder for the user's stack — so it generates verifier code for whatever they run. Keeps setup fully in-app.

mattgwagner: Screenshot 2026-06-16 at 10 50 03@2x

mattgwagner: @choyiny LMK if there's other changes requested here.

choyiny (approved):

Issue #151, opened 2026-06-20
Retry mechanism for outbound sends via downstream providers (Queues)


mattgwagner:
**Problem**

Provider sends are a single synchronous attempt in the request path. When a provider rejects transiently (Cloudflare Email Sending over quota, a 429/5xx, or a network blip), the send just fails — there's no re-attempt, and today the failure is also mis-surfaced (#150). A retry mechanism would re-attempt transient failures with backoff and only surface terminal failures to the user.

The stack already has the substrate: a Cloudflare Queue (`EMAIL_QUEUE` → `saasmail-sequence-emails`, consumer `max_retries: 3`) plus a `queue()` consumer and an hourly `scheduled()` cron (used for sequencing). So this extends an existing pattern rather than introducing new infra.

### A) Async-first — all sends go through a queue
Send/reply routes enqueue an outbound job, persist `sent_emails` as `queued`, and return `202`. A consumer attempts the provider call; on transient failure it calls `message.retry({ delaySeconds })` (native backoff) up to `max_retries`; on exhaustion the job dead-letters and the row becomes `failed`; on success it becomes `sent` with the provider id.
- **+** Single code path; native retry/backoff/dead-letter queue; decouples the user request from provider latency; smooths quota spikes.
- **−** Sending becomes eventually-consistent (UI shows `queued` then transitions via refetch/realtime); biggest change; latency-sensitive sends (transactional/OTP, template "send test") lose synchronous semantics.

### B) Hybrid — sync first attempt, queue only on failure  *(recommended)*
Keep today's synchronous send. On success the row is `sent` immediately (unchanged UX). On a **transient** failure, enqueue a retry job and mark the row `retrying`; the consumer retries with backoff; terminal failure → `failed`.
- **+** Preserves the fast happy path (most sends succeed on the first try and show immediately); pays the async cost only on failure; smallest blast radius; pairs naturally with #150 (a failed send becomes `retrying`, not a phantom `sent`).
- **−** Two code paths; still needs a status lifecycle and transient-vs-permanent error classification.

### C) Cron re-scan — no new queue
Persist `failed` rows with `attempts`/`next_retry_at`; the existing hourly cron re-attempts and gives up after N.
- **+** Simplest; reuses the cron + D1; no new binding.
- **−** Hourly granularity is too coarse for "over quota for a few minutes"; no native backoff/dead-letter; bespoke polling logic.

## Cross-cutting concerns (any approach)
[...]

Issue #150, opened 2026-06-20
Failed sends (provider error, e.g. Cloudflare over quota) still appear as sent messages


mattgwagner:
**Summary**

When the email provider rejects a send (e.g. Cloudflare Email Sending over quota), the API reports the failure but a `sent_emails` row is still persisted and shows in the conversation as if it sent.

**Steps to reproduce**
1. Configure the Cloudflare Email Sending provider and exhaust the daily quota (or otherwise force the provider to reject the send).
2. Compose and send a message from the UI.
3. Observe: the composer closes without an error and the message appears in the thread.

**Expected:** A rejected send is not shown as a sent message; the user sees that it failed.

**Actual:** The message is stored and rendered identically to a successful send, and no error surfaces in the composer.

**Root cause**
- Providers catch failures and return `{ id: null, error }` rather than throwing (`worker/src/lib/email-sender.ts` — `CloudflareSender.send`, ~L154).
- The send (`worker/src/routers/send-router.ts:312`) and reply (`:564`) routes — plus the template test-send (`worker/src/routers/email-templates-router.ts:414`) — insert the `sent_emails` row regardless of `result.error` (flagging `status: "failed"`) and return **HTTP 201**.
- Listing queries don't filter on `status`, and the emails router doesn't expose `status`, so `failed` rows render like successful ones.
- The composer (`src/pages/ComposeModal.tsx:163`) treats the 201 as success and never inspects `result.status`, so the `catch` path never runs.

**Suggested directions** (whichever fits the project's intent)
- Don't persist a `sent_emails` row when the provider errors — or exclude `failed` from the listing queries.
- Return a non-2xx for a failed send, and/or surface `status` in the emails API so the client can mark/hide failures.
- Have the composer/reply UI inspect `status` and show a failure instead of closing as success.

Happy to take the PR if you're open to a direction.

mattgwagner:
I think, at minimum, we should indicate visually on the message that it didn't send successfully. 

I hit my Cloudflare quota overnight and a background job kept trying to resend every 5 minutes... so some of my users show as "receiving" 800+ emails, when in reality, none of them even left SaaSMail.

choyiny:
Looks like you have very high volume!

A visual indicator is definitely good - alongside with #151 I believe an outbox pattern will be fitting.

PR #154, opened 2026-06-21
feat(emails): visual indicator for failed sends (#150)

mattgwagner:
First chunk of #150 — a clear visual indicator when a message didn't actually send.

Sends rejected by the provider (e.g. Cloudflare over quota) are already persisted with `status='failed'` (`send-router.ts`), but that status was never exposed by the API or rendered, so a failed send looked identical to a sent one.

**Changes**
- Surface `status` on the emails API: added to `EmailSchema`, the `by-person` conversation endpoint (sent rows carry their status, received are null), and the single-email endpoint.
- Render a **"Failed to send"** badge + subtle red tint on the message in `MessageBubble` (covers both chat and thread modes) and flip the `ThreadMessage` context-card badge to a red "Failed to send".

Scoped deliberately small: just the indicator, **no filter / list view** — that belongs with the broader outbox/retry work (#151), which will also move failed sends into a dedicated surface. This is the interim visibility fix.

**Test plan**
- `yarn tsc --noEmit` clean; `yarn test` 394 passing (added a case asserting the conversation endpoint returns `status:"failed"` for a failed sent message, null for received).

### DIFF
--- a/src/components/MessageBubble.tsx
+++ b/src/components/MessageBubble.tsx
@@ -1,6 +1,13 @@
 import { useState } from "react";
 import { sanitizeEmailHtml } from "@/lib/sanitize-html";
-import { Link2, Maximize2, Paperclip, Trash2, UserPen } from "lucide-react";
+import {
+  AlertTriangle,
+  Link2,
+  Maximize2,
+  Paperclip,
+  Trash2,
+  UserPen,
+} from "lucide-react";
 import CcChips from "@/components/CcChips";
 import type { Email } from "@/lib/api";
 import { copyMessageLink, messageDomId } from "@/lib/message-link";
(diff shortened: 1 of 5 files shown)


choyiny:
Thanks so much for the quick turnaround - I was going to do it this weekend. Would it be possible to add a manual retry (like iMessage's failed to send, click to retry) in scope for this PR?

mattgwagner:
One caveat worth calling out before adding this: failed sends don't currently persist their attachments. On a provider error the code skips attachment storage (`send-router.ts:329`), and the bytes only ever existed in the original request — so there's no way to recover them on retry, and we can't even detect that a failed send had any.

A manual retry in this PR would therefore re-send the message **body + recipients** only, silently dropping any attachments. Full-fidelity retry needs the bytes staged *before* the first send attempt, which is exactly the outbox staging in #151.

So I'd suggest body+recipients retry here, and close the attachment gap with the outbox work. Want me to proceed on that basis?

mattgwagner:
Thinking about it more, I'd rather not ship a retry that silently drops attachments — that's a worse surprise than no retry button. I'll keep this PR scoped to the visual indicator, and we can add click-to-retry as part of the outbox work in #151, where the bytes are staged before the first attempt so retry is lossless.

choyiny (approved):

PR #163, opened 2026-06-22
refactor(email): split email-sender into providers folder

mattgwagner:
## What & why

`worker/src/lib/email-sender.ts` had grown to ~360 lines holding shared types, helpers, five provider classes, and the factory in one file. This splits it into a focused `email-sender/` folder so each provider is its own file and adding the next one is a small, reviewable diff.

```
worker/src/lib/email-sender/
  index.ts          # createEmailSender factory + re-exports (public surface)
  types.ts          # SendEmailAttachment, SendEmailParams, SendEmailResult, EmailSender
  shared.ts         # parseFrom(), toBase64()
  providers/
    resend.ts  cloudflare.ts  bavimail.ts  noop.ts  demo.ts
```

**Pure code move — no behavior change.** The `./email-sender` import path is unchanged (it resolves to `index.ts`), and `index.ts` re-exports every previously-exported symbol (`createEmailSender`, the type interfaces, and `BavimailSender`), so no callers or tests change.

## Checklist
- [x] `yarn tsc --noEmit` clean
- [x] `yarn test` — 394 passed / 1 skipped, unchanged
- [x] No schema change → no migration
- [x] No user-visible change → no CHANGELOG entry
- [ ] e2e — n/a (no UI/API behavior change)

This is a preparatory refactor; a follow-up PR adds a Postmark provider on top of it.

PR #164, opened 2026-06-22
feat(email): add Postmark send provider

mattgwagner:
## What & why

Adds **Postmark** as an outbound email provider alongside Cloudflare Email Sending, Resend, and Bavimail.

`PostmarkSender` POSTs to `https://api.postmarkapp.com/email` with an `X-Postmark-Server-Token` header. It maps `Reply-To` to Postmark's dedicated `ReplyTo` field and routes any other headers through the `Headers` array, base64-encodes attachments into `Attachments`, and treats both non-2xx responses and `200 + non-zero ErrorCode` as failures. It's fetch-based with an injectable `fetchFn` (so tests never touch `globalThis.fetch`), mirroring `BavimailSender`.

Selected via the `POSTMARK_API_KEY` secret. Runtime precedence: **Bavimail > Postmark > Resend > Cloudflare Email Sending**.

## Stacked on #163

> ⚠️ This PR is stacked on #163 (the `email-sender/` folder refactor) and includes its commit. Please merge #163 first; I'll rebase this onto `main` so it shows only the Postmark commit.

## Changes
- `worker/src/lib/email-sender/providers/postmark.ts` — new `PostmarkSender`.
- `types.ts` union + `index.ts` factory branch & re-export.
- Tests: provider selection (+ precedence), header/cc/attachment mapping, error paths, `maxAttachmentBytes` — appended to `email-sender.test.ts`.
- Docs: README (provider matrix, prerequisites, secrets, `.dev.vars`), `wrangler.jsonc.example`, `.dev.vars.example`, the onboarding skill (Option D), and `CHANGELOG.md`.

## Checklist
- [x] `yarn tsc --noEmit` clean
- [x] `yarn test` — 405 passed / 1 skipped (+11 Postmark tests)
- [x] No schema change → no migration
- [x] CHANGELOG updated under `[Unreleased]`
- [x] Docs updated (README, examples, onboarding)
- [ ] e2e — n/a (no UI/API contract change)

### DIFF
--- a/.claude/skills/saasmail-onboarding/SKILL.md
+++ b/.claude/skills/saasmail-onboarding/SKILL.md
@@ -16,7 +16,7 @@ Read `wrangler.jsonc.example` in the project root so you know the exact shape of
 Set expectations up front, so the user knows what they're committing to:
 
 - **~30–40 minutes** total; most of the wait is DNS propagation, not typing.
-- **Two decisions**: which domains will be used, and which outbound email provider (Cloudflare Email Sending, Resend, or Bavimail).
+- **Two decisions**: which domains will be used, and which outbound email provider (Cloudflare Email Sending, Resend, Bavimail, or Postmark).
 - **Three manual Cloudflare-dashboard steps** (Email Routing per inbound domain, Email Service per send-from domain, and checking the deployed worker).
 - **Cost**: a Cloudflare **Workers Paid plan (~$5/mo)** is required. saasmail uses Queues, which aren't on the free plan. Email Routing is free; Cloudflare Email Sending is usage-based.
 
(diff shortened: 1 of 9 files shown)


mattgwagner:
Heads-up while this is fresh: `BavimailSender` has the **same latent bug** the fix here addresses. It also takes `private fetchFn: typeof fetch = fetch` and calls `this.fetchFn(...)`, so the un-injected production default would throw `Illegal invocation` in the Workers runtime the moment a real Bavimail send runs (its unit tests all inject a mock fetch, so it's currently green but untested on that path).

We hit exactly this with Postmark in production — every send silently failed before reaching the API. I've deliberately left `bavimail.ts` untouched in this PR to keep the diff focused, but you'll likely want the same `fetch.bind(globalThis)` treatment there before anyone relies on Bavimail. Happy to send a separate one-line PR if useful.

mattgwagner:
I've been running this in one of my three production instances for the 4 days and it's working nicely.

mattgwagner:
@choyiny rebased this with latest changes, should be good to go now.

choyiny (approved):

PR #179, opened 2026-07-10
feat: outbox pattern with retry for outbound sends (#151)

### LINKED ISSUE #151: Retry mechanism for outbound sends via downstream providers (Queues)

choyiny:
Closes #151.

Implements the outbox pattern per the maintainer direction on the issue: every email about to be sent is stored in a write-ahead table, the existing hourly cron processes the outbox periodically (approach C), and the outbox is a tab in the application.

## How it works

- **Write-ahead outbox** — a new `outbox_emails` table (migration `0030`). `sendViaOutbox` (`worker/src/lib/outbox.ts`) inserts a row before every provider call, attempts inline (happy-path UX unchanged), then resolves: success/suppressed → row deleted (`sent_emails` stays the history); transient failure → row stays `pending` for retry; permanent failure → terminal `failed`.
- **All four send paths** go through it: compose, reply, template send-test, and sequence steps. On a transient sequence-step failure the queue message is acked and the outbox owns the retry (`sequence_emails` gains a `retrying` status that blocks enrollment completion until resolved).
- **Error classification** — `SendEmailResult.error` now carries `transient: boolean`. Providers with an HTTP status use 429/5xx vs 4xx; Cloudflare binding errors are pattern-matched (quota/rate/timeout/network); unknown defaults to transient so a blip is never silently dropped.
- **Hourly retry** — the existing cron also runs `processOutbox`: atomic conditional-UPDATE claim (no double-send between cron and manual retry), suppression re-checked on every attempt, attachments reloaded from R2, the original `Message-ID` reused for idempotency, and terminal failure after **24 total attempts** (~24h — covers Cloudflare's daily quota reset).
- **API + UI** — `/api/outbox` (list/count/retry/cancel, inbox-scoped), an **Outbox** page with per-row *Retry now* / *Cancel* and a sidebar badge, and an amber *Retrying* indicator on sent messages alongside the existing red *Failed* badge from #150.

## Notes

- No wrangler config changes needed — reuses the existing cron and bindings; provider failover stays out of scope per the issue.
- Migration is hand-written (`0030_outbox_emails.sql` + journal entry) since `drizzle-kit generate` is broken on this repo (see `migrations/README.md`).
- Known trade-off: an uncatchable worker kill between provider accept and outbox delete yields an at-least-once resend an hour later; `Message-ID` reuse is the mitigation.

## Testing

- 455 passed / 1 skipped (`yarn test --run --no-file-parallelism`; parallel mode is pre-existing flaky), `yarn tsc --noEmit` clean, `yarn build` clean, migration applies to local D1.
- New coverage: classification per provider, outbox lifecycle (write-ahead → delete/pending/failed, throw cleanup), processor (claim race, attempt exhaustion, mid-retry suppression, sequence resolution + enrollment completion, Message-ID reuse), router (scoping 403s, 404s, keyset pagination incl. tied timestamps, mid-claim retry refusal), plus route-level integration tests for compose/template/sequence failure paths.


choyiny:
Addressed a 10-finding deep code review (8-angle fan-out + adversarial verification) in 533ac7b:

**Behavioral fixes (each with a regression test):**
1. **Queue-redelivery idempotency** — a D1 error after `sendViaOutbox` handed a sequence step to the outbox used to make the redelivered message re-send with a fresh Message-ID (duplicate email). `processSequenceEmail` now detects the existing outbox row, repairs the bookkeeping (missing `sent_emails` row + step → `retrying`), and acks without re-sending.
2. **Cancel guard** — `DELETE /api/outbox/:id` now refuses (409) a pending row whose send is in flight (mirrors the retry guard), so a cancel can no longer be silently overridden by a concurrent delivery.
3. **Enrollment completion** — a step that fails permanently on the first inline attempt now falls through to `completeEnrollmentIfDone` instead of leaving the enrollment `active` forever (matches the outbox terminal path).
4. **Classification** — `PERMANENT_RE` now catches real Postmark hard rejects ("recipient … marked as inactive", "sender signature not confirmed") that previously defaulted to transient and burned the full 24-attempt budget.
5. **Transient-resolution cool-down** — a 60s `nextRetryAt` buffer closes the ms-wide window where a concurrent claim could resolve against a `sent_emails` row the caller hadn't inserted yet.

**Hardening/docs:**
- Cron phases decoupled: a sequence-dispatch error no longer skips the outbox pass (`.catch(...).then(processOutbox)`).
- `processOutbox` scan bounded (`OUTBOX_BATCH_LIMIT = 200`) with claimed/fetched logging.
- The `status: "retrying"` response value is now documented in the OpenAPI schemas and the use-saasmail skill.
- Mid-retry suppression semantics (`sent_emails` → failed, step → suppressed) documented in-code as intentional.
- Removed a 7×-duplicated identity ternary (`outcome === "sent" ? "sent" : outcome` → `outcome`).

Full serialized suite: 460 passed / 1 skipped; tsc and build clean.

PR #180, opened 2026-07-10
fix: send manually composed emails as transactional

choyiny:
## Problem

Emails composed in the app's compose box are 1:1, human-written messages, but they were being sent as **marketing** email. The frontend `sendEmail()` posted to `/api/send` without a `transactional` flag, which defaults to `false`, so composed emails received:

- `List-Unsubscribe` and `List-Unsubscribe-Post` headers
- an auto-appended "Unsubscribe" footer in the HTML/text body

Replies (`POST /api/send/reply/{emailId}`) were already always transactional; only the compose path was affected.

## Fix

`sendEmail()` in `src/lib/api.ts` now opts into `transactional: true`. The backend `/api/send` route already accepts and honors this flag, so no backend change is needed. Transactional sends skip the unsubscribe headers/footer and bypass the suppression list — consistent with how replies already behave.

The `/api/send` HTTP API **default is unchanged** (`false`), so programmatic/marketing callers are unaffected. Sequence and template sends correctly remain marketing.

## Note

Composed emails now bypass the suppression list (the standard transactional semantic, matching replies). If someone previously unsubscribed, you can still send them a manually-composed 1:1 email.

## Verification

- `yarn tsc --noEmit` passes
- Existing `send-helper` tests already cover the transactional-vs-marketing branching in `send.ts`; no backend logic changed

### DIFF
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -347,7 +347,10 @@ export async function sendEmail(data: {
 }): Promise<{ id: string; attachmentIds: string[]; status: string }> {
   const { files = [], ...payload } = data;
   const fd = new FormData();
-  fd.append("payload", JSON.stringify(payload));
+  // Manually composed emails are 1:1 transactional messages: no unsubscribe
+  // footer or List-Unsubscribe headers, and they bypass the suppression list
+  // (mirrors replies, which are always transactional).
+  fd.append("payload", JSON.stringify({ ...payload, transactional: true }));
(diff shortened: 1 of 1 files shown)


mattgwagner:
Darn, that one is on me. Missed that when did we did that transactional flag.

choyiny:
> Darn, that one is on me. Missed that when did we did that transactional flag.

No problem - now it's caught!

PR #184, opened 2026-07-12
feat(api): register send payload schemas in OpenAPI

mattgwagner:
## Summary
- Register `SendEmailSchema`, `CcEntry`, and `ReplyEmailSchema` under `components.schemas` via `openAPIRegistry.register` (the batch `registerComponent('schemas', {...})` form does not register Zod schemas).
- Document `ReplyEmailSchema` fields on the reply route payload description with an example.
- Include `transactional` in the send payload example; field was already in the Zod schema but absent from `/doc`.

## Test plan
- [x] `npx vitest run --config vitest.config.test.ts worker/src/__tests__/openapi-send-schemas.test.ts`
- [x] `npx vitest run --config vitest.config.test.ts worker/src/__tests__/send-router.test.ts`
- [x] `npx prettier --check` on changed files
- [ ] `curl -s '$URL/doc?cb=1' | jq '.components.schemas | keys'`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- OpenAPI `/doc`: register `SendEmailSchema`, `CcEntry`, and `ReplyEmailSchema` under `components.schemas` (including `transactional` and reply payload fields).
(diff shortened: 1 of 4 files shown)


choyiny (approved):
Reviewed: changes look good. Approving.

PR #187, opened 2026-07-12
fix: clear stale Reply-To when reassigning a message to a new person

mattgwagner:
## Summary
- Strip the inbound `Reply-To` from `raw_headers` when a received message is re-attributed to a different person, so `people.email` stays the canonical reply target
- Return `replyTo: null` from the single-email endpoint when the stored `Reply-To` matches the attributed sender (covers messages reassigned before this fix)
- Improve the reassign modal prefill to fall back to `fromAddress` when there is no distinct `Reply-To`

## Test plan
- [x] `npm test -- --run reassign-person.test.ts emails-router.test.ts`
- [ ] Reassign a contact-form message (with `Reply-To`) to a different person; confirm the modal no longer pre-fills the old address
- [ ] Reply after reassignment still routes to the new person's email

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/components/ReassignPersonModal.tsx
+++ b/src/components/ReassignPersonModal.tsx
@@ -29,9 +29,10 @@ export default function ReassignPersonModal({
   const [submitting, setSubmitting] = useState(false);
   const [error, setError] = useState<string | null>(null);
 
-  // On open, prefill the address from the message's inbound Reply-To when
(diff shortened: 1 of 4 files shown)


choyiny (approved):
Reviewed: changes look good. Approving.

PR #232, opened 2026-08-05
ci(release-drafter): apply version labels on fork pull requests

Waiel5:
## Summary

`Check PR labels` fails any PR that carries none of `major`/`minor`/`patch`. Those labels are supposed to come from release-drafter's autolabeler — but it runs on `pull_request`, where the `GITHUB_TOKEN` is **read-only for pull requests opened from a fork**. So the labeler cannot write a label on exactly the PRs that need one, and an outside contributor (who also cannot self-label) is blocked until a maintainer intervenes by hand.

I hit this opening the other PRs in this batch, so it seemed worth fixing first.

## Changes

- Move autolabelling into its own workflow on `pull_request_target`, which runs in the base repository's context and can write labels. This is the same event `Check PR labels` already uses. The job **checks out no PR code** and only reads PR metadata through the API, so untrusted code from a fork is never executed.
- Release Drafter keeps drafting on `push` to main, now with `disable-autolabeler: true` so the two roles do not overlap.
- Add the version-bump checkboxes to the PR template. The autolabeler matches `/\[x] Patch/`, `/\[x] Minor/` and `/\[x] Major/` against the body and the template contained none of them, so there was nothing to match even where it had permission. Change-type checkboxes are included too, so merged PRs file under the right release-notes heading.

## Release impact

- [x] Chore

## Test plan

- [x] `yarn test`
- [x] Workflow YAML parsed and validated (triggers/jobs assert as expected)
- Behaviour itself is only observable on a real fork PR — this PR is one, so the checkboxes above should exercise it once merged.

## Notes for reviewers

`pull_request_target` deserves scrutiny, so to be explicit about the safety argument: the risk with that event is checking out and executing PR-head code with a privileged token. This job has no `actions/checkout` step at all and runs only `release-drafter` against the API. Permissions are `contents: read` + `pull-requests: write` and nothing else.

I split it from `release-drafter.yml` rather than adding a trigger there, so the release-drafting job keeps `contents: write` on `push` only and the fork-facing job never gets it.

## Checklist

- [ ] Updated docs — n/a

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,6 +8,26 @@
 
 -
 
+## Release impact
+
+<!--
+Tick exactly one version bump. This applies the label that "Check PR labels"
+requires — without it the check stays red. Tick the change type too, so the
+release notes file this PR under the right heading.
+-->
+
+- [ ] Patch — bug fix or internal change, no new behaviour
+- [ ] Minor — new backwards-compatible behaviour
+- [ ] Major — breaking change
+
+<!-- Change type (optional, categorises the release notes): -->
+
+- [ ] Feature
+- [ ] Bug Fix
+- [ ] Refactor
+- [ ] Documentation
+- [ ] Chore
+
 ## Test plan
 
 <!-- How did you verify this works? Include commands run and/or scenarios exercised. -->
--- a/.github/workflows/pr-autolabel.yml
+++ b/.github/workflows/pr-autolabel.yml
@@ -1,0 +1,36 @@
+name: Autolabel PRs
+
+# Applies the major/minor/patch label that "Check PR labels" requires, based on
+# the checkboxes in the PR body (see .github/release-drafter.yml autolabeler).
+#
+# This is deliberately separate from the Release Drafter workflow, which runs on
+# `push` to main. On `pull_request`, the GITHUB_TOKEN is read-only for pull
+# requests opened from a fork, so the autolabeler silently cannot apply a label
+# and every outside contributor's PR stays red until a maintainer labels it by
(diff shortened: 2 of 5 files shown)


Waiel5:
Heads-up on how this PR (and the four others I opened) currently present, since the page is misleading rather than informative:

**"All checks have failed" is one check, and it is this bug.**

Exactly one check-run exists on the commit — `Check PR labels` — and it fails with:

```
Error! This pull request does not contain any of the valid labels: ['major', 'minor', 'patch']
```

That is the problem this PR fixes, and it cannot fix it for itself: `pull_request_target` reads its workflow definition from the **base branch**, so whatever is on `main` is what runs. The fix only takes effect once merged.

**Nothing else has run.** `Test`, `e2e`, `Format`, `CodeQL` and `Release Drafter` are all sitting at `action_required` — GitHub holds `pull_request` workflows from a first-time fork contributor until you approve them. They produced no check-runs at all, which is why the rollup is 1/1 rather than 1/6.

**The failing check is displayed as `test`.** That is `pr-labels.yml`'s job name, not the test suite — which, as above, has not executed. I have pushed a second commit renaming that job to `verify-label`, since a missing label presenting as "test failed" on a PR where no test ran seems worth removing. Happy to drop that commit if you would rather keep the job name stable.

I can't clear the red myself — `Waiel5` has no push access, so it cannot apply a label. It needs either a label from you or this merged.

Sorry for the noise on what should be a small PR; it seemed better to explain than to leave five red PRs looking broken.

choyiny:
I'm fine with labelling these myself at the moment, the CI failure for checking PR labels is intentional so that it allows me to review first.

PR #233, opened 2026-08-05
fix(attachments): scope attachment reads to the caller's allowed inboxes

Waiel5:
## Summary

`GET /api/attachments/{id}` and `/{id}/inline` looked an attachment up by id and served it, without ever consulting `allowedInboxes`. **Any authenticated user could read any attachment in the deployment given only its id**, including from inboxes they hold no permission on. It appears to be the one place in the API where inbox scoping is absent rather than enforced.

Found while reading the API surface; I have not tried it against a live deployment beyond the added tests.

## Changes

- Both routes resolve the owning inbox through the attachment's message and apply the same predicate the email detail routes use (`isInboxAllowed`).
- The owning inbox depends on `kind`. An inbound attachment belongs to `emails.recipient`; a **sent** one belongs to `sent_emails.from_address` — its `to_address` is the external correspondent, not an inbox anyone holds permission on. Scoping sent attachments by recipient would have been wrong in both directions.
- An attachment the caller may not read answers **404, not 403**, matching what `emails-router` already returns for the same reason: a 403 confirms the id exists.
- Inline responses stop advertising `Cache-Control: public`. These are authenticated mailbox bytes, and `public` licenses shared caches and proxies in front of the worker to store one caller's mail and serve it to another. Now `private`, keeping the immutable long-lived caching in the requesting browser where it is useful and harmless.

## Release impact

- [x] Bug Fix

## Test plan

- [x] `yarn test worker/src/__tests__/attachments-router.test.ts` — 11 passed
- [x] Verified the new tests **fail against the previous handler** (4 failures), so they actually cover the gap rather than just passing
- [x] Ran the full blast radius — attachments, emails, conversations, people, send, delete-email, delete-person, openapi-email-schemas — 82 passed

## Notes for reviewers

Worth knowing why a full suite never caught this: **every existing test in `attachments-router.test.ts` authenticates as an admin**, and `resolveAllowedInboxes` short-circuits to `{ isAdmin: true }` for admins, skipping all scoping. The path was structurally invisible to the tests that existed. The added cases run as a member with a single inbox grant and cover both routes and both kinds, plus an admin case to confirm nothing narrowed for them.

Happy to split the `Cache-Control` change into its own PR if you'd rather keep this to the authorization fix.

## Checklist

- [ ] Updated docs — n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 
-- Attachment reads are scoped to the caller's allowed inboxes. `GET /api/attachments/{id}` and `/{id}/inline` looked the attachment up by id alone and never consulted `allowedInboxes`, so any authenticated user could read any attachment in the deployment — including from inboxes they hold no permission on — given only its id. Both routes now resolve the owning inbox through the attachment's message and apply the same check the email detail routes make: an inbound attachment is scoped by `emails.recipient`, a sent one by `sent_emails.from_address` (the external `to_address` is not an inbox anyone holds). An attachment the caller may not read answers `404`, not `403`, so the response does not confirm that the id exists. Every existing test in the suite authenticated as an admin, for whom `resolveAllowedInboxes` short-circuits, which is why this went unnoticed.
-- Inline attachments are no longer served with `Cache-Control: public`. The responses are authenticated mailbox content, and `public` licenses shared caches and proxies in front of the worker to store them and serve them to a different caller. They are now `private`, keeping the immutable long-lived caching in the requesting browser only.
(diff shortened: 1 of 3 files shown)


Waiel5:

Locally: 11 passed on `attachments-router`, and the 4 new authorization cases fail against the previous handler. Full blast radius — attachments, emails, conversations, people, send, delete-email, delete-person, openapi-email-schemas — 82 passed.

choyiny:
Thanks for your contribution! For code changes as simple as this - I recommend trimming down on the AI generated in-line comments. Too long for such as straight forward implementation.

Waiel5:
Gotcha on those AI comments, to be honest I was going to go through a couple of them on here just to ensure everything was good.

choyiny (approved):

choyiny:
Thanks for your contribution!

PR #235, opened 2026-08-05
fix(notifications): stop notifying users who may not read mail

Waiel5:
## Summary

Inbound fanout picks recipients from `inbox_permissions` plus admins and applies no further check, so two kinds of user keep receiving notifications after being cut off from the API:

- A **banned** user. `/mcp` refuses them outright as "account suspended", but their subscriptions are still delivered to.
- A user with **no registered passkey**. `requirePasskey` 403s every `/api/*` request they make, so they cannot open the message — yet the notification still arrives.

This matters because the payload is not a bare "you have mail" ping. It carries the sender's name, the subject, and a 140-character body preview (`do/notifications.ts`), so delivering it to a user the API refuses hands them mail content through a side channel.

## Changes

- `filterNotifiableUsers` applies both conditions between target computation and delivery, so `computeFanoutTargets` stays pure and separately testable.
- Ban check mirrors the MCP handler's (`banned && (!banExpires || banExpires > now)`), so an expired ban still notifies.
- The passkey condition is skipped in development via `isDevEnvironment`, matching `requirePasskey` and the MCP handler, so local dev and e2e still receive notifications.
- Logs a count when recipients are skipped.

## Release impact

- [x] Bug Fix

## Test plan

- [x] `yarn test` on notification-fanout, do-notifications, notifications-router, inbound-forward — 56 passed
- [x] Verified the new tests fail when the filter is stubbed to a pass-through (3 failures), so they assert real behaviour

## Notes for reviewers

The passkey half is the debatable one, so I want to be explicit about the reasoning rather than slip it in. My first instinct was that a passkey-less user is merely mid-onboarding and should still be told mail arrived. What changed my mind is `require-passkey.ts`'s own stated purpose — *"so passkey registration is actually required to access data"* — and a subject line plus body preview is data. It is the same argument `mcp/http.ts` makes for re-checking passkeys on bearer requests instead of exempting itself. If you read that differently I am happy to gate only on `banned` and drop the passkey condition; it is a two-line change.

One test detail worth noting: the tests pass `env` explicitly rather than relying on the ambient bindings, which set `DISABLE_PASSKEY_GATE="true"`. Under those the passkey branch never executes, so a test written against them would silently assert nothing.

## Checklist

- [ ] Updated docs — n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- Notifications are no longer delivered to users who may not read mail. Inbound fanout selected recipients from `inbox_permissions` plus admins and checked nothing else, so a **banned** user — refused everywhere else as "account suspended" — and a user with **no registered passkey** — refused by `requirePasskey` on every `/api/*` call — both kept receiving WebSocket frames and Web Push carrying the sender's name, the subject, and a 140-character body preview. That is mail content, not merely a signal that mail arrived, so it made "passkey registration is required to access data" false for anyone holding a subscription. Both conditions are now applied before delivery, with the passkey condition skipped in development exactly as `requirePasskey` and the MCP handler skip it.
+
 ### Fixed
 
 - README: correct OpenAPI doc URLs (`/doc` and `/swagger-ui`, not `/api/doc`) and OpenAPI version (3.0).
--- a/worker/src/__tests__/notification-fanout.test.ts
+++ b/worker/src/__tests__/notification-fanout.test.ts
@@ -93,15 +93,8 @@ describe("computeFanoutTargets", () => {
   });
 });
 
-/**
- * `computeFanoutTargets` decides who is *interested*; this decides who is
- * *permitted*. The notification payload carries sender, subject and a body
- * preview, so a user the API refuses must not be sent one.
- *
- * The env is passed explicitly rather than taken from the ambient test
- * bindings, which set DISABLE_PASSKEY_GATE="true" — under those the passkey
- * branch never runs, so a test relying on them would assert nothing.
- */
+// Env is passed explicitly: the ambient test bindings set
+// DISABLE_PASSKEY_GATE="true", under which the passkey branch never runs.
 describe("filterNotifiableUsers", () => {
   const GATE_ON = { DISABLE_PASSKEY_GATE: "false" } as any;
   const GATE_OFF = { DISABLE_PASSKEY_GATE: "true" } as any;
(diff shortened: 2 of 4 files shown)


Waiel5:

Locally: 56 passed across `notification-fanout`, `do-notifications`, `notifications-router` and `inbound-forward`, and the 3 new cases fail when the filter is stubbed to a pass-through.

choyiny:
There isn't a path to banning a user at this moment, but I think there's an opportunity of making it so that when a user is removed, all the subscriptions should be removed as well?

Waiel5:
Agreed, and I think that is the better shape — removing a user should take their subscriptions with them. Worth noting `DELETE /api/admin/users/{id}` already gets that for free: `push_subscriptions.user_id` and `expo_push_subscriptions.user_id` both cascade, so the rows go with the user today.

The case this PR is about is the one where the user is still *present* but should not be receiving mail content. Two of those exist right now:

- **No registered passkey.** `requirePasskey` refuses every `/api/*` call, so the user cannot read a message in the app — but fanout still pushed them the sender, the subject and a 140-character body preview. That makes "passkey registration is required to access data" untrue for anyone holding a subscription.
- **Banned.** You are right that nothing in the UI sets it, but the column is there via better-auth's admin plugin and `resolveOAuthPrincipal` already refuses a banned user everywhere else. It seemed worth having fanout agree with the rest of the codebase rather than being the one path that does not check.

So I would frame it as: cascade-on-delete handles the deleted user, and this handles the present-but-refused one. Happy to drop the ban half if you would rather not carry a check for a state nothing sets yet — the passkey half is the one I actually hit.

Waiel5:
Pushed a fix to this branch — I introduced a bug in it and would rather flag it than let it through.

`filterNotifiableUsers` passed the full candidate list straight into `inArray(users.id, ...)`. `MAX_ADMIN_FANOUT` caps the admin half, but the `inbox_permissions` half is uncapped, and D1 rejects a query carrying more than 100 bound parameters. So on an inbox with enough members the query throws, the `ctx.waitUntil` catch in `email-handler.ts` swallows it as a warning, and *nobody* gets notified for that message — silently, while the inbound mail still succeeds.

That is worst on exactly the deployments this PR is for: a collaborative inbox with a real team.

Both `inArray` sites now chunk at 90, concurrently, with the ban and passkey predicates applied to the union afterwards — same candidates in, same eligible users out, and at or below 90 the emitted SQL is unchanged.

The ceiling is verified rather than assumed: Cloudflare documents 100, and probing the workerd D1 the suite runs against gives n=100 OK / n=101 throws. Two tests, one of which asserts that the chosen size genuinely exceeds the ceiling, so if that limit ever rises the test fails loudly instead of quietly covering nothing. Red-before-green checked by reverting the chunking and confirming the failure is the D1 error itself.

I also bumped that catch from `warn` to `error` and gave it the inbox and email id. Control flow is unchanged — still non-fatal — but a total fanout failure should not read the same as "no recipients".

choyiny:

PR #236, opened 2026-08-05
fix(emails): register the bulk route before /{id} so it is reachable

Waiel5:
## Summary

`PATCH /api/emails/bulk` is declared after `PATCH /api/emails/{id}`. Hono matches routes in registration order, so the parameterised route wins: a bulk request binds `id: "bulk"`, runs the single-email handler, finds no email with that id and returns 404. Nothing is marked, and the bulk handler has never run — it is unreachable dead code.

## Changes

- Move the bulk route's registration above `/{id}`. The handler itself is unchanged.

## Release impact

- [x] Bug Fix

## Test plan

- [x] `yarn test worker/src/__tests__/emails-router.test.ts` — 27 passed
- [x] Verified the two new behavioural tests **fail against the previous ordering**

## Notes for reviewers

The handler was already scoped correctly — it skips ids the caller has no inbox permission for rather than failing the whole request — so making it reachable does not widen access. There is a test covering that, since a route that has never executed deserves one before it starts.

Reordering a route table is exactly the kind of fix that repairs one path by breaking another, so there is also a test asserting a single-email `PATCH /api/emails/{id}` still reaches the id handler.

Since this route has never worked, nothing can depend on its current 404 behaviour — but it does mean any client that was silently getting 404s here will now start actually marking mail read.

## Checklist

- [ ] Updated docs — n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


Waiel5:

Locally: 27 passed on `emails-router`, and the 2 new behavioural cases fail against the previous route ordering.

PR #237, opened 2026-08-05
fix(auth): keep the pending OAuth request across sign-in

Waiel5:
## Summary

When a client starts an OAuth authorization and the browser has no session, the provider redirects to `/login` with the pending request in the query string, signed and time-bounded by the server. Both sign-in handlers then run `window.location.href = "/"`, discarding it. The user lands on the inbox and **the client that started the flow waits for a callback that never arrives.**

This is the path every third-party and native client takes on first connect, since such a client never has a browser session to begin with.

Two reasons it went unnoticed: every existing OAuth test signs in *before* calling authorize (`runFlow` in `mcp-helpers.ts`), so the logged-out branch has no coverage; and `/login` and `/consent` sit outside `AuthGuard`, so nothing else routes through it either.

## Changes

- Sign-in hands the pending parameters back to `/api/auth/oauth2/authorize` instead of navigating to `/`. Applies to both the passkey and the password path.
- An ordinary sign-in with no pending request still goes to `/`, unchanged.

## Release impact

- [x] Bug Fix

## Test plan

- [x] `yarn test worker/src/__tests__/oauth-logged-out-authorize.test.ts` — 4 passed
- [x] `npx playwright test e2e/specs/oauth-login-continuation.spec.ts` — 2 passed
- [x] **Verified the e2e test fails against the previous handler**, with `sign-in dropped the pending authorization and landed on /`

## Notes for reviewers

**On the redirect being safe.** It is not an open redirect. The destination is always this deployment's own `/api/auth/oauth2/authorize`, never a URL read out of the query; only the parameters are carried over, and the server re-verifies its own `sig` and `exp` before acting on them. The gate is the presence of both `client_id` and `sig`, so an arbitrary query string does not trigger it.

**On the two test layers.** The failure spans the browser and the server, so I covered both. The worker test pins the contract the page depends on — authorize redirects an unauthenticated request to the login page with signed, expiring parameters; replaying them with a session resumes rather than restarts; and no code is ever issued to an unauthenticated caller. Those pass on `main` too, which is the point: they document a previously untested branch rather than chase the bug. The e2e spec is the one that actually fails without the fix.

The e2e spec registers its client through RFC 7591 dynamic registration, exactly as an unknown client would, so it exercises the real path rather than a fixture.

**Local environment note, in case it helps anyone else:** running the e2e suite on macOS needed `CHOKIDAR_USEPOLLING=1` — miniflare's file watcher hits `EMFILE: too many open files` against `node_modules` otherwise. CI is unaffected. Happy to add that to the README if you'd find it useful.

## Checklist

- [ ] Updated docs — n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- README: correct OpenAPI doc URLs (`/doc` and `/swagger-ui`, not `/api/doc`) and OpenAPI version (3.0).
+- **Signing in no longer abandons a pending OAuth authorization.** When a client starts an authorization while the browser has no session, the provider redirects to `/login` carrying the request as signed, time-bounded query parameters — but both sign-in paths then navigated to `/`, discarding it. The client waited for a callback that never arrived. Sign-in now hands those parameters back to `/api/auth/oauth2/authorize` and the flow continues to consent. This is the path every third-party client takes on first connect, since it always arrives logged out; the existing OAuth tests all authenticate before calling authorize, so the branch had no coverage.
 
 ### Added
 
--- a/e2e/specs/oauth-login-continuation.spec.ts
+++ b/e2e/specs/oauth-login-continuation.spec.ts
@@ -1,0 +1,116 @@
+// e2e/specs/oauth-login-continuation.spec.ts
+// Covers: a third-party client starting an OAuth authorization while logged
+// out, and the flow surviving the trip through the login page.
+//
+// This is the path every native and third-party client takes on first connect,
+// and the only one where the browser has no session yet. The worker-side tests
+// all sign in before calling authorize, so without this the login page could
+// drop the pending request and nothing would notice.
+import { test, expect } from "../fixtures/test";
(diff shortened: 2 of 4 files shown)


choyiny:
I would appreciate reproduction steps before attempting a PR for it, thanks!

Waiel5:
Repro, minimal:

1. Deploy an instance and sign out of the dashboard entirely.
2. From any OAuth client, hit `GET /api/auth/oauth2/authorize?client_id=...&redirect_uri=...&response_type=code&code_challenge=...&code_challenge_method=S256`.
3. The provider sees no session and redirects to `/login`, carrying the request back as signed, time-bounded query parameters.
4. Sign in.
5. You land on `/`. The authorize request is gone, and the client waits on a callback that never arrives.

`src/pages/LoginPage.tsx` navigated to `/` unconditionally on success, in both the passkey and the password path, discarding the parameters the provider had just handed it. The fix hands them back to `/api/auth/oauth2/authorize` so the flow continues to consent.

This is the path every third-party client takes on first connect, since it always arrives logged out. The existing OAuth tests all authenticate *before* calling authorize, which is why the branch had no coverage — the PR adds an e2e that fails without the change.

choyiny:

PR #239, opened 2026-08-05
feat(notifications): let OAuth clients open the realtime stream

Waiel5:
## Summary

`GET /api/notifications/stream` rejects any upgrade whose `Origin` is not in `TRUSTED_ORIGINS`. A native client has no browser origin to send, so the realtime stream is unreachable for exactly the clients the OAuth surface exists to serve.

**Stacked on #238** — it contains that PR's commit, so review this one's diff after it lands (or read the top commit only).

## Changes

The `Origin` check now follows the credential it protects rather than the route: unchanged for session cookies, skipped for OAuth.

## Release impact

- [x] Feature

## Test plan

- [x] `yarn test worker/src/__tests__/notifications-router.test.ts` — 10 passed, including a real 101 upgrade
- [x] The three existing 403 cases still pass unchanged

## Notes for reviewers

**The security argument, since this relaxes a control you added deliberately** (`CHANGELOG.md` records it as CSWSH defence).

Cross-Site WebSocket Hijacking works because a browser attaches the session cookie to a cross-origin handshake *by itself*: an attacker's page opens the socket and the victim's credential rides along. `Origin` is the right control there precisely because a page cannot forge it.

A bearer credential is not ambient. It has to be set explicitly on the request, and no attacker page can cause a victim's access token to be attached to a socket it opened — so the attack the check prevents cannot occur for these callers, while the check itself is fatal to them.

**Scoped to OAuth on purpose.** The same argument applies to `sk_` API keys, and I nearly extended it there — but the suite asserts a 403 for API-key callers today, and changing a tested contract is not needed to unblock a native client. Left alone rather than quietly widened. Say the word if you would rather it were consistent.

**One correction included:** this file's header said the 101 upgrade was not covered because it needed a real WebSocket-capable client. The `NOTIFICATIONS_HUB` binding is in the test config and the handshake does complete under miniflare — the new cases assert it, and the comment is updated.

## Checklist

- [ ] Migration — n/a
- [x] Updated `CHANGELOG.md`
- [ ] Docs — the route's OpenAPI description is updated to explain the split


choyiny:

PR #240, opened 2026-08-05
feat(api): advertise an API version and capabilities on /api/config

Waiel5:
## Summary

The web UI is deployed from the same commit as the worker, so it never has to ask what the server supports. A third-party or native client is a separate artifact talking to whatever version an operator happens to be running, and has no way to find out — its only option is to attempt a capability and interpret the failure, which is indistinguishable from a bug, a misconfiguration, or a revoked grant.

That matters most because the failure arrives late: a client can complete dynamic registration, walk the user through an authorization it cannot finish, and only discover on its first API call that this deployment does not accept tokens there — after the user has already approved something.

**Stacked on #238 and #239.**

## Changes

`GET /api/config` gains `apiVersion` and a `capabilities` object (`oauthApi`, `oauthStream`).

## Release impact

- [x] Feature

## Test plan

- [x] `yarn test worker/src/__tests__/bootstrap-capabilities.test.ts` — 6 passed
- [x] `openapi-bootstrap` and `openapi-doc` still pass

## Notes for reviewers

**Additive and safe for existing consumers.** `src/lib/branding.tsx` type-guards `passkeyRequired` and `brandName` individually and ignores everything else, so the extra fields cannot disturb it.

`apiVersion` is deliberately separate from the package version, which moves for reasons a client does not care about — a dependency bump should not tell a client the contract changed.

The convention I would suggest, and have written into the schema comment: add a capability key when one lands, never remove one, since a client may branch on its absence.

## Checklist

- [ ] Migration — n/a
- [x] Updated `CHANGELOG.md`
- [ ] Docs — the fields are documented in the OpenAPI schema


choyiny:

PR #241, opened 2026-08-05
feat(notifications): deliver to native devices via Expo push

Waiel5:
## Summary

Web Push is predicated on a browser `PushSubscription` — a push-service URL plus the P-256 keypair the `PushManager` generated, which the worker encrypts each payload to. React Native has no ServiceWorker and no `PushManager`, so a native client has no such keypair to offer; it has an opaque `ExponentPushToken` and nothing else. This is a second transport for the same event, not a variant of the existing one. **The Web Push path is unchanged.**

**Stacked on #238, #239, #240.**

## Changes

- New `expo_push_subscriptions` table (migration `0033`, purely additive) and `POST`/`DELETE /api/notifications/expo/subscribe`.
- Delivery branches by transport inside `NotificationsHub`.
- Notification previews off by default, `PUSH_PREVIEWS="true"` to opt in.
- VAPID no longer required for deployments serving only native clients.

## Release impact

- [x] Feature

## Test plan

- [x] `yarn test worker/src/__tests__/expo-push.test.ts` — 18 passed
- [x] Full notification surface — expo-push, do-notifications, notifications-router, openapi-notifications, web-push, notification-fanout, openapi-doc — **45 passed**
- [x] `yarn db:generate` produced `0033`; verified it contains **no** `ALTER`, `DROP` or `RENAME`
- [x] `prettier --check .` clean

## Notes for reviewers

**Why a separate table.** Sharing `push_subscriptions` means making `p256dh` and `auth` nullable, and SQLite cannot drop a NOT NULL constraint in place — it needs a create-copy-drop-rename of a live table that also carries a unique index and a foreign key. That is the riskiest migration in the repo for the least benefit. The new table is one `CREATE TABLE` and two indexes.

**Why installation id rather than token.** Expo rotates tokens on reinstall and on restore to a new handset. Keying on the token accumulates a dead row per rotation, and each one still looks live until something tries to send to it. Re-registering updates the row in place; the version counter moves only when the token actually changed, so it tracks rotations rather than app launches.

**Previews default to off, and I'd argue that's the right default for this product.** A native payload passes through Expo *and* then Apple or Google before it reaches a lock screen — a different exposure from Web Push, which is encrypted end-to-end to the browser. By default the notification says only that mail arrived and carries opaque routing ids. `PUSH_PREVIEWS="true"` opts in per deployment.

That matters more than usual here because **the send cannot be authenticated**. APNs and FCM credentials live on the Expo project that built the app, not on a self-hosted deployment, so the server POSTs to `exp.host` unauthenticated and Expo routes by token — which makes a token a bearer capability for anyone who obtains one. Tokens are stored server-side and never returned by any route, and with previews off a stolen token cannot be used to *read* anything. If your Expo project ever enables enhanced push security, unauthenticated sends stop working and this needs an access token; that is a deliberate either/or rather than something I've papered over.

**Deliberately best-effort.** No durable job ledger, no cross-request retry. A lost notification means the user opens the app and sees the mail — the WebSocket stream and pull-to-refresh already cover that, and building a delivery-guarantee system for a wake-up ping seemed like the wrong trade. It chunks to Expo's 100-message limit, stops on a 429 rather than hammering a rate limiter, and prunes only on `DeviceNotRegistered`. Receipts (Expo's delayed second signal) are not polled — worth adding if delivery problems show up, but tickets catch the case that matters, which is a dead token.

One thing worth knowing if you review the code: **Expo reports per-message failure inside an HTTP 200**, so the ticket array has to be read rather than the status code trusted. There is a test for exactly that.

**Two fixes fall out:**
- The VAPID guard sat at the top of the delivery path, so an unset key skipped *every* transport rather than just Web Push.
- TTL is 24h on the native path. The web path hardcodes 60s, which is aggressive for a browser and simply wrong for a phone that is asleep or underground.

`via: "push"` still means push was *attempted*, not that a send succeeded — I changed that briefly and the existing test caught it.

## Checklist

- [x] Added a migration (`yarn db:generate`) — `0033`, additive only
- [x] Updated `CHANGELOG.md`
- [x] Documented `PUSH_PREVIEWS` in `wrangler.jsonc.example`


choyiny:
I'm not certain what this PR is trying to achieve - we already support notifications to iOS and Android devices using Web Push: see https://caniuse.com/push-api for compatibility.

Waiel5:
this is for the mobile app I was adding. You can check it here https://github.com/Waiel5/saasmail-mobile.

All things are basically to ensure it would be ready for a mobile app, currently still working on perfecting it

choyiny:
Appreciate you working on a mobile app for SaaSmail! This is super cool. At this time, we do not want to store expo tokens within saasmail.

Wondering what the benefits are of having a mobile app vs just installing the web app to the Home Screen on iOS and Android

choyiny:

PR #242, opened 2026-08-05
feat(api): list the inboxes the caller may send from

Waiel5:
Stacked on #238 — the last commit here is the only new one; the scope-policy entry belongs with the route it governs, so it ships alongside rather than after. Review the tip commit.

## The gap

`GET /api/admin/inboxes` is admin-only, and it was the only inbox list in the API.

A member therefore had no way to discover a `fromAddress` that `POST /api/send` would accept. A client could hardcode one, or build a From picker by guessing and let the user find out from a 403. Neither is a contract, and it makes composing from a non-admin account impossible to implement correctly — which is how I hit it, building a From picker in a native client.

## The route

`GET /api/inboxes` returns the caller's own sending identities: every inbox for an admin, the assigned ones for a member.

It is built from the same universe the admin route builds — received recipients `UNION` configured sender identities — and narrowed with `inboxScopeSql`, the same resolver `POST /api/send` enforces with. That equivalence is the whole point of the route, so it is asserted directly rather than assumed: one test sends from a listed address and an unlisted one and checks the pair does not 403 / does 403. The two cannot drift into a picker that offers addresses which fail.

The response carries `displayName`, `displayMode` and `signatureHtml` — what a client needs to render and label a From row. It deliberately omits two fields the admin route returns:

- `forwardTo` is a standing relay destination and admin configuration.
- `assignedUserIds` discloses which other users hold an inbox.

## Tests

10 tests, `worker/src/__tests__/inboxes-router.test.ts`.

The member path is set up explicitly in every scoping test. `createTestUser` defaults to `role: "admin"` and `resolveAllowedInboxes` short-circuits on `isAdmin`, so a suite that omits that setup asserts an unrestricted caller sees everything — green, and exercising none of the scoping.

The empty grant gets its own test. A member with no assignments must get `[]`, and both plausible shortcuts fail: an inlined `IN ()` is a SQLite syntax error (500), and skipping the filter when the list is empty scopes nothing and returns every inbox on the deployment. `inboxScopeSql` is the one place that renders it as a false predicate, which is why the query goes through it rather than a hand-rolled `IN`.

Also covered: case-insensitive matching against grants stored before insert-time canonicalization, the `chat` default for an inbox with no `sender_identities` row, 401 unauthenticated, and the scope classification.

## Scope

Classified `email:read`. It discloses addresses the caller already controls and sends nothing, and a read-only client still needs it to label a message with the inbox that received it. Unclassified routes are denied outright by #238, so leaving it out would let an OAuth client read mail but never find a From address.

## Verification

```
Test Files  3 passed (3)
     Tests  51 passed (51)
```
(`inboxes-router`, `oauth-api-access`, `inbox-permission-enforcement`.) `yarn tsc --noEmit` clean.


choyiny:

PR #244, opened 2026-08-05
feat(api): draw the OAuth admin boundary around durable privilege

Waiel5:
Stacked on #238 — review the last three commits. Depends on that PR's scope policy existing at all.

## Why

#238 refused nine things to OAuth bearer callers. Building a native admin client against it, the line turned out to be drawn by **category** rather than by **damage**, and to be wrong in both directions:

- `DELETE /api/admin/invites/{id}` is denied, and it *destroys* a capability. It was only ever caught by a `(\/|$)` prefix over the whole invites surface — an accident of pattern shape.
- `DELETE /api/blocklist/mail` is wide open at `admin:manage`, and it hard-deletes mail and R2 objects across every inbox in the deployment, irreversibly, with no dry-run.

## The rule

One test, stated in the file header: **does this create a durable privilege, or open a standing exfiltration channel?** Durable meaning it survives revoking the client and expiring the token.

Where an operation failed the test only because of *how much* it could grant, it is clamped rather than closed — the clamp is what removes the durable property:

| Operation | Now |
|---|---|
| `POST /api/admin/invites` | Open, clamped to `role: "member"` + pinned `email` + ≤7 days. One call yields a single-target, short-lived, non-escalating credential. |
| `PATCH /api/admin/users/{id}/role` | Open for **demotion only**. Demoting cannot escalate. |
| `PATCH /api/admin/inboxes/{email}` | Open, except setting a non-empty `forwardTo`. Clearing stays open, so a client keeps a kill switch it can never use to arm. |
| `PUT /api/webhook` | Same split on `url`. Secret rotation now works (see below). |
| `PUT .../assignments` | Open **shrink-only** — may remove an assignment, never add one. |
| `DELETE /api/admin/invites/{id}`, `POST /api/webhook/test`, `GET`+`DELETE /api/oauth-apps` | Open. |
| all of `/api/api-keys`, `/api/user/passkeys`, `/api/auth/*`, `GET /api/admin/invites` | **Still denied.** |

`/api/api-keys` stays shut permanently: an `sk_` key resolves as `authMethod:"apiKey"`, which skips the scope middleware *and* `requirePasskey`, has no expiry, and is not revoked by `DELETE /api/oauth-apps/{clientId}`. One call would buy permanent unscoped access.

`GET /api/admin/invites` stays shut because it spreads `...row` and returns the raw, still-usable `token` of every live invite — including admin-role ones a human created in a browser.

## Guards are allowlists

`BODY_GUARDS` refuses any body key it does not classify, mirroring `classifyRoute`'s trailing `return denied`. A test enumerates each route's zod schema and asserts every key is either allowed or explicitly refused, so a field added next year is refused to tokens by default. Guards read `authMethod` and are invisible to session and API-key callers, asserted directly.

## Two holes found by review, closed here

**Assignments were a composition hole.** Every step permitted, and together durable: mint a member invite pinned to an address you control (the clamp allows exactly that, and the 201 body carries the token), redeem it through the public accept route, then hand that member every inbox. The resulting access outlives revoking the client. Assignments are now shrink-only — the analogue of demote-only.

**Webhook secret rotation was dead permission.** The guard marks `secret: "free"`, but the route made it impossible: sending the current URL back with a new secret is refused, omitting the URL failed validation (`url` was required), and a blank URL wiped the configuration before `secret` was read. `url` is now optional and absent means "leave the destination alone" — which cannot arm anything, so the guard still covers its path. Found by an agent writing the client's webhook screen, which went looking for the control the policy said it could draw and reported the permission as dead rather than shipping a button that silently deleted the webhook it claimed to re-key.

## Verification

```
Test Files  3 passed (3)
     Tests  77 passed (77)
```
Non-vacuity checked by mutation: making the invite clamp permissive and moving the invites deny rule one line below the general admin rule made 7 tests fail. Full suite 827 passed / 1 skipped (needs `--fileParallelism=false` here; the default deadlocks ~60 workerd children at 0% CPU, pre-existing).

## Known, not fixed here

`/api/blocklist` is mounted without `requireAdmin` in `index.ts`, so this policy is stricter for OAuth than the session path is for anyone. Worth a separate fix; flagging rather than widening this diff.


choyiny:

PR #247, opened 2026-08-06
feat(api): report what each domain still needs to receive mail

Waiel5:
Independent of my other PRs — branches from `main`.

## Why

saasmail has no concept of a domain: inboxes are keyed by full address. So an operator who has just pointed a new domain at their deployment has no way to see whether it actually works, and nothing tells them what is wrong when mail bounces.

That is not hypothetical — on my own instance Email Routing was enabled and `ready`, the MX records were correct, and every message still bounced `550 5.1.1` because the catch-all rule was `disabled, action: drop`. Nothing in saasmail could have told me that.

## What

`GET /api/domains`, admin-only. Domains are derived from the addresses that already exist — `DISTINCT emails.recipient` UNION `sender_identities.email`, split at the `@` — so there is **no new table and no migration**. Per domain: inbox count, message count, and a live DNS status.

The DNS part is the useful half. A Worker cannot do raw DNS, but it can query Cloudflare's DoH resolver, so MX, SPF and DKIM are each resolved independently and reported as `cloudflare` / `elsewhere` / `none` / `unknown`, with the exact records still missing.

Two things I want to draw attention to because getting them wrong is worse than not shipping the feature:

**SPF is merged, never appended.** Only one SPF record per name is legal, so a domain already carrying `v=spf1 include:_spf.google.com ~all` gets back that record with the Cloudflare include spliced in before the terminal `all`, marked `action: "replace"`. An earlier revision emitted the canonical standalone value, which an operator following it literally would have published as a second record — breaking their outbound deliverability. Cloudflare's own docs say the same: *"If you have existing SPF records, merge them"*.

**DKIM's value is `null`, deliberately.** The selector `cf2024-1._domainkey` is documented and checked, but the `p=` key is generated per zone, so the response reports the record as required with a null value and a note telling the operator to copy it from the dashboard. A wrong DKIM value is worse than an absent one.

`unknown` is never a verdict about the zone. A resolver timeout, a non-2xx, an unparseable body and SERVFAIL all read as "could not check", and NXDOMAIN reads as "no record" — those are different answers and are not conflated. The three checks resolve independently, so a failed TXT lookup does not suppress what MX established.

## Tests

23 in `worker/src/__tests__/domains-router.test.ts`. `fetch` is stubbed and the helper throws on any domain not declared, so an accidental real lookup fails loudly rather than passing by luck.

Two defects found while writing them, both fixed here:

- the SPF include check was a bare `includes()`, so `include:_spf.mx.cloudflare.net.evil.example` read as ours and the domain showed green. It now matches whole terms with the qualifier stripped, and the same predicate decides both the status and whether a merge is needed, so the two cannot disagree.
- `txtValue` claimed to concatenate long TXT strings and actually stripped quotes, which corrupts any record split across chunks.

```
Test Files  77 passed (77)
     Tests  771 passed | 1 skipped (772)
```
`yarn tsc --noEmit` clean.

## Note on scope

`worker/src/lib/oauth-scope-policy.ts` does not exist on `main` — it arrives with #238. When these meet, `GET /api/domains` needs an `admin` rule with `requiresAdminRole`, since unclassified routes are denied to bearer callers. Happy to fold that in whenever #238 lands.


choyiny:

PR #248, opened 2026-08-06
fix(ui): put mark-all-read in the same place on both thread types

mattgwagner:
Opening a thread, I could never predict where the mark-all-read control would be. It turns out it isn't just left-vs-right — the two thread surfaces disagree on four things at once:

| | Person thread | Group conversation |
|---|---|---|
| Position | Left | Right |
| Row | Its own full-width strip below the inbox tabs | Inside the header row |
| Style | Borderless text button | Outlined button |
| Label | `Mark all in support as read` + count | `Mark all read`, no count |

So it's a re-find every time, and the strip on the person view costs a row of vertical space on every thread that has unread mail.

Before and after

**What this does:** extracts the affordance into a single `MarkAllReadButton` and renders it from the same slot on both surfaces — the trailing action cluster of the header row. One component means the two can't drift apart again. The person view loses the extra strip entirely.

**What this deliberately does not change:** scope. The person view still marks only the *active inbox tab* read; the group view still marks the whole thread. Those are correct for their respective surfaces — a person has multiple inbox tabs, a group conversation is keyed to one inbox. The button label no longer tries to carry the scope; the tooltip does (`Mark all 7 unread in support@example.com as read`).

One honest note on "same place": on the group view the button is flush right (it's the only header action); on the person view it sits first in the existing `[mark read] [add to sequence] [block]` cluster, so it isn't pixel-identical. I kept that order rather than pushing it past `Block` — benign action before destructive one. Same row, same cluster, same style, which is what made it findable.

Screenshots are from `seeds/demo.sql`, not real mail.

Typecheck clean, `yarn test` green (76 files / 746 tests).

### DIFF
--- a/src/components/MarkAllReadButton.tsx
+++ b/src/components/MarkAllReadButton.tsx
@@ -1,0 +1,43 @@
+import { CheckCheck } from "lucide-react";
+
+interface MarkAllReadButtonProps {
+  /** Number of unread received emails in scope. Renders nothing at 0. */
+  unreadCount: number;
+  /**
+   * What "all" covers, for the tooltip — e.g. `support@example.com` on a
+   * person's inbox tab, or the conversation's inbox on a group thread.
+   * The button label stays the same either way; only the tooltip narrows it.
+   */
+  scopeLabel: string;
+  onMarkAllRead: () => void;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


choyiny (approved):
thanks, this works for me!

PR #253, opened 2026-08-10
fix(blocklist): mark matching unread mail read when blocking

mattgwagner:
## Summary
- Blocking a sender (or domain) now marks that sender’s unread received mail as read and recomputes `people.unread_count`.
- That keeps the nav Inbox badge aligned with the people list (which already hides blocked senders), without teaching `/api/stats` about the blocklist.
- Idempotent `POST /api/blocklist` also re-runs the mark-read so any unread that arrived before the rule (or while it was briefly removed) heals.
- Follow-up to [@choyiny’s suggestion](https://github.com/choyiny/saasmail/pull/253#issuecomment-5247903963).

## Test plan
- [x] `vitest` — blocklist create marks email + domain matches read; leaves unrelated unread alone; stats unread drops accordingly
- [ ] Block a sender with unread mail → badge decreases; person disappears from inbox
- [ ] Re-POST the same rule → still 200, leftover unread (if any) cleared

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Inbox list: hydrate group participants/CC after pagination so `GET /api/people/grouped` no longer 500s on mailboxes with 50+ group threads (D1's 100 bound-parameter cap). Stats still counted unread while the people list failed empty.
+- Blocklist: mark matching unread mail as read when a rule is created, so the nav unread badge cannot stick on senders the inbox list has hidden. Migration `0033` clears existing blocked unread counts.
 - README: correct OpenAPI doc URLs (`/doc` and `/swagger-ui`, not `/api/doc`) and OpenAPI version (3.0).
 
 ### Added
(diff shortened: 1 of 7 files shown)


choyiny:
I feel like it may make more sense to set everything to read when we blocklist a certain person. Thoughts?

mattgwagner:
Agreed — that’s the better place to fix it.

Updated this PR to mark matching unread mail as read when a block rule is created (and on idempotent re-POST), then recompute `people.unread_count`. Dropped the `/api/stats` filter so the badge stays a plain unread count again.

choyiny:
Thanks for changing this up - if we can remove the other changes, that will be helpful. Then a simple migration script will be able to remove the counts without having to add a one-time feature to the codebase.

mattgwagner:
Trimmed this down to just the mark-on-create behavior, plus migration `0033` to clear existing blocked unread counts. Dropped the shared not-blocked SQL helper refactor and the idempotent re-POST heal path.

choyiny (approved):

choyiny:
Regenerated migration `0033` through the drizzle-kit CLI instead of hand-authoring it (pushed as `38595d1`).

Since this is a data backfill (UPDATE-only, no schema change), `drizzle-kit generate` on its own emits nothing ("No schema changes, nothing to migrate"). The right CLI path for a data migration is `drizzle-kit generate --custom --name=mark_blocked_senders_read`, which produces a properly-journaled empty migration to paste the SQL into.

What changed:
- Added the missing `migrations/meta/0033_snapshot.json` — the hand-authored version had no snapshot, which desyncs the drizzle-kit journal.
- Journal `when` now reflects the real generation time.
- **SQL body is byte-for-byte identical** to the previous version; verified it still applies cleanly to local D1 (`3 commands executed successfully`).

PR #294, opened 2026-09-22
feat(gmail): sync Workspace mailboxes on a cron

choyiny:
> **⚠️ This branch is STACKED on #292 and #293.** Both are merged into it, so the diff against `main` contains all three slices. Review this one against `2c01807` (the second merge commit) or wait until the parents land. Nothing here conflicts with them.

## Summary

Third of four slices adding Gmail / Google Workspace inbox sync. **Mail now actually arrives.** Each connected Google mailbox is polled every 15 minutes, new messages are pulled via the Gmail API and fed through saasmail's existing ingestion path — so they land on the customer timeline exactly as Cloudflare-routed mail does, with the same person matching, conversation grouping, attachments, webhooks and notifications.

- **Schema** — `sender_identities` gains `source` / `gmail_account_id` / `gmail_group_address`; `emails` gains Gmail's message and thread ids; `gmail_accounts` gains `last_gap_at`. Existing rows default to `source = "cloudflare"` and are untouched. (migrations `0036`, `0037`)
- **Gmail API client** — `history.list` paging and `messages.get(format=raw)`, returning the same RFC822 bytes Cloudflare hands the Email Worker, which is what lets both sources share one parser.
- **Sync engine** — a resumable history-cursor loop with per-account isolation, record-boundary truncation, and a bounded per-run message cap.
- **Admin API** — map an inbox to a connected account; the connected-accounts route now surfaces `last_gap_at`.
- **Cron** — `0 * * * *` → `*/15 * * * *`, running alongside the existing sequence dispatch and outbox processing.

## Personal mailboxes only — Google Group routing was withdrawn

The original design inferred a destination inbox from message headers, so one "collector" mailbox could serve several Groups. **Four successive matchers were each broken by adversarial review**, ending in two Criticals: `List-ID` is fully sender-forgeable, and `Delivered-To` is forgeable via a duplicate-header collapse in the parser. Header-derived routing hands an attacker the choice of which customer's timeline their message lands on.

So this slice ships personal mailboxes only, where an account maps 1:1 to one inbox and **no header is consulted at all**. The `gmail_group_address` column is retained for a later slice, but nothing can configure it: the admin API returns 400 with *"Google Group routing is not supported yet. Only personal mailboxes can be mapped to an inbox in this release."* A field an operator can set that silently does nothing is worse than an absent one.

A follow-up slice should key on something a sender cannot set — a distinct collector address per group, or a server-side Gmail label.

## The bug this PR almost shipped

Moving the cron to `*/15` quietly broke the **outbound** retry budget. `MAX_OUTBOX_ATTEMPTS = 24` is a count of *cron ticks*, not a duration — its own comment read "with the hourly cron this is ~24h — enough to ride out Cloudflare's daily send-quota reset". At four times the cadence that budget burned in ~6h, so quota-blocked mail would have been marked permanently `failed` roughly 18 hours before the quota reset.

Caught by the whole-branch review, not by any per-task one: the cron is shared and `outbox.ts` belonged to no task in the plan. Now `96`, with a comment that states the coupling explicitly. I audited the rest of the codebase — it is the only retry budget denominated in ticks.

## Failure behaviour

The organising question for every review here was *can this lose mail, or silently stop delivering it?* The final review could not construct any sequence — truncation, poison message, malformed history, re-seed, restart, overlapping ticks, backwards cursor — that advances the cursor past unprocessed mail.

- **Per-account isolation.** One failing mailbox never stops the others.
- **Stall, never drop.** A malformed history record throws rather than being skipped; a poison message stops that run with the cursor at the last completed record; an unmapped or ambiguously-mapped account pauses without consuming any history. Every one is visible via `lastError`.
- **Resumable truncation.** A run capped at 50 messages advances only to the last *fully completed* history record, so the remainder drains on the next tick. An end-to-end test pushes 60 messages through two runs and asserts all 60 arrive.
- **Sync gaps are recorded.** If a mailbox goes unpolled for ~a week Gmail expires the cursor and sync re-seeds from the present — mail inside that gap is unrecoverable. `last_gap_at` records it permanently and the admin route exposes it.
- **Skipped:** `SENT`, `DRAFT`, `TRASH`, `SPAM`.

## Testing

`yarn tsc -b` → zero errors under `worker/`. `yarn format:check` clean. All nine touched test files pass together (gmail sync/api/routing/mapping/cron, both admin routers, outbox processor).

**I have not observed the full suite green on this machine, and I'm not going to claim otherwise.** It has been running at load 26–102 on 14 cores from unrelated work, and `vitest-pool-workers` uses a 5-second test timeout; full runs failed 42–46 of 92 files. Classified: 46 timeouts, **zero "no such column"** (the signal that would matter, since this slice adds columns in two migrations), and 3 assertion failures — all in files this branch never touches, one of which is a cross-test state-bleed flake independently diagnosed during slice 1. Those files pass in isolation. Please confirm on CI.

Two standing caveats from earlier slices still apply: the 6 pre-existing frontend `src/` type errors, and CI's typecheck step running `yarn tsc --noEmit`, which is a no-op in this repo.

## Known gaps

- **The cron wiring is not covered end to end.** Both tests call `syncAllGmailAccounts` exactly as `index.ts` calls it, not through the `scheduled` export — `cloudflare:test` does not resolve here (verified by probe: `Cannot find package 'cloudflare:test'`; this project uses `cloudflarePool()` rather than `defineWorkersConfig`). **Deleting the Gmail-sync block from `index.ts` would fail no test.** The failure mode is loud — nothing ever syncs — and closing it means restructuring the Vitest config for all 92 files.
- No uniqueness or foreign-key check on `gmail_account_id`; an ambiguous mapping stalls visibly rather than misrouting.
- Deferred to slice 4: `SENT`-label ingestion (replies typed in Gmail don't appear yet), read-state write-back, sending, and any UI.
- Worth a separate PR against #293's file: `email-parser.ts` collapses repeated headers to the last value, dropping extra `Delivered-To:` lines.


choyiny:
Update on the test-suite caveat in the description.

The PR says I had not observed the full suite green on this machine. I have now: **92/92 test files pass**, run against this branch's head with the machine at load ~12 on 14 cores rather than the 26–102 it had been sitting at while the slice was being built.

So the 42–46 file failures described in the description were entirely environmental — `vitest-pool-workers` uses a 5-second test timeout and the box was heavily oversubscribed by unrelated work. The classification in the description holds up: all timeouts, zero schema errors, and the handful of assertion failures were cross-test state bleed in files this branch never touches.

No code changed to achieve this; it is the same commit (`5780cef`), just run on a quieter machine. Leaving the original caveat in the description rather than editing it, so the record of what I could and could not verify at the time stays intact.

